PR #170, opened 2024-03-20
Align parsing of credential_response_encryption to draft 13

### LINKED ISSUE #143: Align parsing of `credential_response_encryption` to draft 13
Draft 13 clarified the meaning of `encryption_required` especially in case is set to `false`.

Currently, our library checks that when this flag is set to false, issuer should not define encryption methods.

Draft 13 clarifies that this is not the case. When setting false the flag it is ok to have encryption methods and it is up to the wallet whether it will request for an encrypted response or not.

In this case, it would be nice to change our code to request encryption from the issuer or add a configuration option

vafeini:
Closes #143 

Draft 13 describes 3 "states" of the issuer's metadata regarding encrypted responses:
1. Encrypted responses not supported (NOT_SUPPORTED)
2. Encrypted responses supported but its up to the requestor to select if issuer's response will be encrypted or not (SUPPORTED_NOT_REQUIRED) 
3. Encrypted responses supported and always required (REQUIRED)

The above is reflected in the sealed hierarchy eu.europa.ec.eudi.openid4vci.CredentialResponseEncryption included in issuer metadata domain object. These states should be taken into account in the ResponseEncryptionSpecFactory passed during the construction on an Issuer. The result of this factory method is validated internaly in the issuer against the 3 cases mentioned above,


src/main/kotlin/eu/europa/ec/eudi/openid4vci/internal/IssuanceServerClient.kt
```diff
@@ -166,18 +165,14 @@ internal class IssuanceServerClient(
...
-            when (issuerMetadata.credentialResponseEncryption) {
-                is CredentialResponseEncryption.NotRequired -> {
+            when (val encryptionSpec = request.encryption) {
+                null -> {
                     val success = response.body<SingleIssuanceSuccessResponse>()
                     success.toDomain()
                 }
 
-                is CredentialResponseEncryption.Required -> {
+                else -> {
```
babisRoutis, line 174: Why use else? 

There are two distinct cases: 
- Required (which corresponds to the code of else)
- SupportedNotRequired 
The later should be governed by a configuration option, IMO

src/main/kotlin/eu/europa/ec/eudi/openid4vci/internal/IssuanceServerClient.kt
```diff
@@ -206,12 +201,12 @@ internal class IssuanceServerClient(
...
         if (response.status.isSuccess()) {
             when (issuerMetadata.credentialResponseEncryption) {
-                is CredentialResponseEncryption.NotRequired -> {
+                is CredentialResponseEncryption.NotSupported -> {
                     val success = response.body<BatchIssuanceSuccessResponse>()
                     success.toDomain()
                 }
 
-                is CredentialResponseEncryption.Required -> {
+                else -> {
```
babisRoutis, line 209: Similar with the previous comment. There are two distinct cases.

src/main/kotlin/eu/europa/ec/eudi/openid4vci/CredentialIssuerMetadataResolver.kt
```diff
@@ -17,13 +17,23 @@ package eu.europa.ec.eudi.openid4vci
 
 import com.nimbusds.jose.EncryptionMethod
 import com.nimbusds.jose.JWEAlgorithm
-import eu.europa.ec.eudi.openid4vci.CredentialResponseEncryption.NotRequired
+import eu.europa.ec.eudi.openid4vci.CredentialResponseEncryption.NotSupported
 import eu.europa.ec.eudi.openid4vci.internal.DefaultCredentialIssuerMetadataResolver
 import java.io.Serializable
 
 sealed interface CredentialResponseEncryption : Serializable {
```
babisRoutis, line 24: Here we need to introduce a data class containing the algs and ecryption method.
Then nest this to SupportedNotRequired and/or Required.

babisRoutis (changes requested):
Please check my comments

babisRoutis (approved):

PR #173, opened 2024-03-22
Introduce Credential Response Encryption policy.

### LINKED ISSUE #171: Allow wallet to define its policy about response encryption
So far the library tries to match the issuer's requirements for the encryption of the response (that carries the issued credential).
In general VCI allows the issuer to chose one of the following options:
- **Not Used** (`credential_response_encryption` is not provided to issuer's metadata)
- **Supported but not required** (`credential_response_encryption` is present and `encryption_required` is `false`)
- **Required** (`credential_response_encryption` is present and `encryption_required` is `true`)

Wallet could also define  a similar policy independently of the issuer's choice. During resolution of a credential offer there could be a step to assure that those two policies are compatible and proceed with the issuance.

 The wallet's policy can be the following:
 - **Required** 
 - **Supported**
 
|                                             |  Required  | Supported  |
|--------------------------------------|---------------|----------------|
| Not used                              |  Boom       |   No encryption   |
| Supported but not required  |  Encrypt   |  Try to encrypt. Fallback to no encryption    |
| Required                              |  Encrypt   |   Encrypt      |

dzarras:
Closes #171


src/main/kotlin/eu/europa/ec/eudi/openid4vci/internal/RequestIssuanceImpl.kt
```diff
@@ -32,29 +32,61 @@ internal class RequestIssuanceImpl(
...
-            is CredentialResponseEncryption.SupportedNotRequired ->
-                if (config.preferEncryptedResponsesWhenSupported) {
-                    val supportedAlgorithmsAndMethods = encryption.encryptionAlgorithmsAndMethods
-                    responseEncryptionSpecFactory(supportedAlgorithmsAndMethods, config.keyGenerationConfig)?.apply {
+            is CredentialResponseEncryption.SupportedNotRequired -> {
+                // Issuance server supports but does not require Credential Response encryption.
+                // Fail in case Wallet requires Credential Response encryption but no crypto material can be generated,
+                // or in case algorithm/method supported by Wallet is not supported by issuance server.
+                val supportedAlgorithmsAndMethods = encryption.encryptionAlgorithmsAndMethods
+                val maybeSpec = responseEncryptionSpecFactory(supportedAlgorithmsAndMethods, config.keyGenerationConfig)
```
babisRoutis, line 56: Please wrap this into `runCatching{ ... }.getONull()` block.
Even the default implementation of `responseEncryptionSpecFactory` raises in some cases exception.

src/main/kotlin/eu/europa/ec/eudi/openid4vci/internal/RequestIssuanceImpl.kt
```diff
@@ -32,29 +32,61 @@ internal class RequestIssuanceImpl(
...
 
             is CredentialResponseEncryption.Required -> {
-                val supportedAlgorithmsAndMethods = encryption.encryptionAlgorithmsAndMethods
-                responseEncryptionSpecFactory(supportedAlgorithmsAndMethods, config.keyGenerationConfig)?.apply {
-                    validate(supportedAlgorithmsAndMethods)
-                } ?: throw IssuerExpectsResponseEncryptionCryptoMaterialButNotProvided
+                // Issuance server requires Credential Response encryption.
+                // Fail in case Wallet does not support Credential Response encryption or,
+                // algorithms/methods supported by Wallet are not supported by issuance server.
+                when (config.credentialResponseEncryptionPolicy) {
```
babisRoutis, line 75: Do we need the `when` at this point?
I don't think so

babisRoutis (changes requested):
@dzarras Can you please tackle the comments?

vafeini:
We need the following 2 unit tests to be added:
- when wallet mandates encrypted responses but issuer does not then excpetion ResponseEncryptionRequiredByWalletButNotSupportedByIssuer is thrown
- when wallet mandates encrypted responses but no cryptographic can be generated WalletRequiresCredentialResponseEncryptionButNoCryptoMaterialCanBeGenerated is thrown

vafeini (dismissed):

babisRoutis (changes requested):
@dzarras Can you please implement in this PR the first solution of
https://github.com/eu-digital-identity-wallet/eudi-lib-jvm-openid4vci-kt/issues/171#issuecomment-2025383657

dzarras:
> * WalletRequiresCredentialResponseEncryptionButNoCryptoMaterialCanBeGenerated

Added.

dzarras:
@babisRoutis 

> Can you please implement in this PR the first solution of [#171 (comment)](https://github.com/eu-digital-identity-wallet/eudi-lib-jvm-openid4vci-kt/issues/171#issuecomment-2025383657)

Done. Please check.

vafeini (approved):

babisRoutis (approved):

PR #203, opened 2024-04-25
Add 'iss' claim to jwt key proof

### LINKED ISSUE #137: Missing issuer claim from Proof of type JWT
In file `RequestIssuanceImpl`

we should add the following

```kotlin
         ProofBuilder.ofType(ProofType.JWT) {
            iss(config.clientId)
            aud(issuerMetadata.credentialIssuerIdentifier.toString())
            publicKey(proofSigner.getBindingKey())
            credentialSpec(credentialSupported)
            nonce(cNonce.value)
            build(proofSigner)
        }
```

That is to add the issuer claim

ydanneg:
According to spec `iss` claim is optional and MUST be a `client_id`.
https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#section-7.2.1.3-2.2.2.1

I didn't make it optional, I used the same pattern as for `c_nonce`, which is also OPTIONAL by spec, but required by the code.

@babisRoutis please check if this PR is valid.

Fixes #137

### DIFF
--- a/src/main/kotlin/eu/europa/ec/eudi/openid4vci/Issuance.kt
+++ b/src/main/kotlin/eu/europa/ec/eudi/openid4vci/Issuance.kt
@@ -281,7 +281,7 @@ interface RequestIssuance {
      */
     suspend fun AuthorizedRequest.NoProofRequired.handleInvalidProof(
         cNonce: CNonce,
-        clientId: String
+        clientId: ClientId,
     ): AuthorizedRequest.ProofRequired
 }
 
--- a/src/main/kotlin/eu/europa/ec/eudi/openid4vci/Issuance.kt
+++ b/src/main/kotlin/eu/europa/ec/eudi/openid4vci/Issuance.kt
@@ -66,7 +66,6 @@ sealed interface AuthorizedRequest : java.io.Serializable {
         override val accessToken: AccessToken,
         override val refreshToken: RefreshToken?,
         val cNonce: CNonce,
-        val clientId: ClientId,
         override val credentialIdentifiers: Map<CredentialConfigurationIdentifier, List<CredentialIdentifier>>?,
     ) : AuthorizedRequest
 }
--- a/src/main/kotlin/eu/europa/ec/eudi/openid4vci/Issuance.kt
+++ b/src/main/kotlin/eu/europa/ec/eudi/openid4vci/Issuance.kt
@@ -66,6 +66,7 @@ sealed interface AuthorizedRequest : java.io.Serializable {
         override val accessToken: AccessToken,
         override val refreshToken: RefreshToken?,
         val cNonce: CNonce,
+        val clientId: String,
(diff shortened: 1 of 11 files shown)


babisRoutis:
Hi @ydanneg 

I understand that this is a fix for #137 .
Is this the case?

babisRoutis:
@ydanneg Can you please run `./gradlew spotlessApply` and then commit & push the formatted code?

ydanneg:
> Hi @ydanneg
> 
> I understand that this is a fix for #137 . Is this the case?

Somehow I missed this issue :) Seems you right, this is the one.

babisRoutis:
As a first comment (haven't check PR in detail), I notice that `clientId` has been added to various classes/methods.
Yet `clientId` is constant known before using the `Issuer`. That is why this attribute can be found already within `OpenId4VCIConfig`. 

So, perhaps, instead of adding the `clientId` we could just make available the configuration (if it is not) where we need to use it

babisRoutis:
> As a first comment (haven't check PR in detail), I notice that `clientId` has been added to various classes/methods. Yet `clientId` is constant known before using the `Issuer`. That is why this attribute can be found already within `OpenId4VCIConfig`.
> 
> So, perhaps, instead of adding the `clientId` we could just make available the configuration (if it is not) where we need to use it

In particular I think that we just need to add clientId as a private property of `RequestIssuanceImpl`

ydanneg:
> As a first comment (haven't check PR in detail), I notice that `clientId` has been added to various classes/methods. Yet `clientId` is constant known before using the `Issuer`. That is why this attribute can be found already within `OpenId4VCIConfig`.
> 
> So, perhaps, instead of adding the `clientId` we could just make available the configuration (if it is not) where we need to use it

thanks for the tip, now it looks much better

babisRoutis:
Dear @ydanneg 


Current PR just reverts the changes that were introduced, which introduces commits not actually needed and in addition there are some clientId(s) introduced that need to be removed.

May I ask you to create a new PR, applying just the change to `RequestIssuanceImpl`?
I think it would more "clean".
What do you think? 

PS : You just removed them :smile:

ydanneg:
> Dear @ydanneg
> 
> Current PR just reverts the changes that were introduced, which introduces commits not actually needed and in addition there are some clientId(s) introduced that need to be removed.
> 
> May I ask you to create a new PR, applying just the change to `RequestIssuanceImpl`? I think it would more "clean". What do you think?
> 
> PS : You just removed them 😄

Can't you just squash it?

ydanneg:
> > Dear @ydanneg
> > Current PR just reverts the changes that were introduced, which introduces commits not actually needed and in addition there are some clientId(s) introduced that need to be removed.
> > May I ask you to create a new PR, applying just the change to `RequestIssuanceImpl`? I think it would more "clean". What do you think?
> > PS : You just removed them 😄
> 
> Can't you just squash it?

nevermind.. i will do a new one :)

ydanneg:
will do another one, with a single commit

PR #204, opened 2024-04-25
fixes #137: add iss to jwt key proof

### LINKED ISSUE #137: Missing issuer claim from Proof of type JWT

ydanneg:
Fixes #137


babisRoutis (approved):

babisRoutis:
@ydanneg Thanks again for the fix

I merged it, leaving aside the considerations expressed in my previous comment, because I have the impression that in most cases the `iss` is required. It is exceptional the case that it is has to be omitted

PR #212, opened 2024-04-30
Reorganize OAUTH2 & Credential Issuer HTTP clients

### LINKED ISSUE #213: Improve package/code structure
HTTP clients for OAUTH2 & Credential Issuer use transfer objects - representing requests / responses.
Some of those transfer objects are in `.internal.format` package whereas others are in `.internal`

We need to improve this and in particular put all related transfer object into the same package

babisRoutis:
Closes #213

This PR re-factors the HTTP clients for placing calls towards the OAUTH2 / OIDC and Credential Issuer servers.

A new package `.internal.http` was added and clients were moved there.
Also in this package Transfer objects (for JSON) were moved


babisRoutis:
@vafeini Perhaps PR is too extensive...
Haven't implemented yet handling encrypted batch response. It should be straightforward

PR #215, opened 2024-05-02
Make credential offer available from Issuer

### LINKED ISSUE #216: Accessing credential offer via `Issuer`
When instantiating `Issuer` using a deep link representing a credential offer, caller has no access to the resolved credential offer.

A solution would be to make the resolved credential offer accessible from the `Issuer`

babisRoutis:
This PR is a small improvement in ergonomics.

Adds to `Issuer` interface a public property for accessing the credential offer for which the Issuer was created.

Closes #216


ydanneg:
@babisRoutis 
can we have a new release please?

PR #225, opened 2024-05-22
Support attestation-based client authentication

babisRoutis:


babisRoutis:
Closing this PR, since main has diverged.

Part of it will be probably included in a new PR

PR #226, opened 2024-05-22
Deferred response encryption

### LINKED ISSUE #227: Support encrypted responses in deferred endpoint
Based on discussion under OpenID4VCI issue https://github.com/openid/OpenID4VCI/issues/286 encrypted responses for deferred issuance requests must be supported the following way:

 
1. If initial issuance request (batch or single) had a deferred response (transaction_id instead of the credential itself) wallet must remember the response encryption information specified in the initial request 
2. When accesing the deferred endpoint with the transaction_id returned by the issuer, wallet must expect that the response should be encrypted based on the encryption information kept from step 1.

vafeini:
Closed #227

### DIFF
--- a/src/main/kotlin/eu/europa/ec/eudi/openid4vci/Issuance.kt
+++ b/src/main/kotlin/eu/europa/ec/eudi/openid4vci/Issuance.kt
@@ -301,7 +301,9 @@ sealed interface DeferredCredentialQueryOutcome {
     data class Issued(val credential: IssuedCredential.Issued) : DeferredCredentialQueryOutcome
 
     data class IssuancePending(
+        val transactionId: TransactionId,
         val interval: Long? = null,
+        val responseEncryptionSpec: IssuanceResponseEncryptionSpec? = null,
     ) : DeferredCredentialQueryOutcome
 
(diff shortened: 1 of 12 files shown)


src/main/kotlin/eu/europa/ec/eudi/openid4vci/Issuance.kt
```diff
@@ -115,10 +115,12 @@ sealed interface SubmittedRequest : java.io.Serializable {
      * If the issuance request was a batch request, it will contain the results of each issuance request.
      * If it was a single issuance request list will contain only one result.
      * @param cNonce Nonce information sent back from the issuance server.
+     * @param responseEncryptionSpec The response encryption information as specified in the issuance request.
      */
     data class Success(
         val credentials: List<IssuedCredential>,
         val cNonce: CNonce?,
+        val responseEncryptionSpec: IssuanceResponseEncryptionSpec?,
```
babisRoutis, line 123: Please remove the `responseEncryptionSpec` attribute.

babisRoutis (changes requested):

PR #228, opened 2024-05-22
Fix authorization server metadata resolution

### LINKED ISSUE #224: oauth-authorization-server related issue,  not able to reach the correct endpoint
Hi, I think there might be an issue that cause reference wallet related to “.well-known/oauth-authorization-server”.
The behaviour of current reference wallet appending “.well-known/oauth-authorization-server” to a url with path components different with what specs have been stated.
Let's say we have https://example.com/path1/path2
Current behaviour of reference wallet is appending it right behind the url like :https://example.com/path1/path2/.well-known/oauth-authorization-server

According to the spec https://www.rfc-editor.org/rfc/rfc8414.html#section-3
“.well-known/oauth-authorization-server” need to be inserted after removing the path components first. e.g : https://example.com/.well-known/oauth-authorization-server/path1/path2

dzarras:
Closes #224

### DIFF
--- a/src/main/kotlin/eu/europa/ec/eudi/openid4vci/internal/DefaultAuthorizationServerMetadataResolver.kt
+++ b/src/main/kotlin/eu/europa/ec/eudi/openid4vci/internal/DefaultAuthorizationServerMetadataResolver.kt
@@ -27,7 +27,6 @@ import io.ktor.client.call.*
 import io.ktor.client.request.*
 import io.ktor.http.*
 import net.minidev.json.JSONObject
-import java.net.URL
 
 internal class DefaultAuthorizationServerMetadataResolver(
     private val httpClient: HttpClient,
@@ -44,13 +43,10 @@ internal class DefaultAuthorizationServerMetadataResolver(
      */
     private suspend fun fetchOidcServerMetadata(issuer: HttpsUrl): Result<CIAuthorizationServerMetadata> =
         runCatching {
-            val url =
-                URLBuilder(issuer.value.toString())
-                    .appendPathSegments("/.well-known/openid-configuration", encodeSlash = false)
-                    .build()
-                    .toURI()
-                    .toURL()
-
+            val url = issuer.metadataUrl("/.well-known/openid-configuration")
+            fetchAndParse(url, OIDCProviderMetadata::parse)
+        }.recoverCatching {
+            val url = issuer.metadataUrl("/.well-known/openid-configuration", false)
             fetchAndParse(url, OIDCProviderMetadata::parse)
         }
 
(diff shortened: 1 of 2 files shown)


src/main/kotlin/eu/europa/ec/eudi/openid4vci/internal/DefaultAuthorizationServerMetadataResolver.kt
```diff
@@ -85,3 +78,23 @@ internal class DefaultAuthorizationServerMetadataResolver(
...
     require(issuer == Issuer(expected.value.toURI())) { "issuer does not match the expected value" }
+
+/**
+ * Gets a well-known metadata URL of an issuer.
+ *
+ * @receiver the Issuer
+ * @param metadata the well-known metadata URL to get
+ * @param specCompliant whether to get a spec-compliant URL or not
+ */
+private fun HttpsUrl.metadataUrl(metadata: String, specCompliant: Boolean = true): Url =
```
babisRoutis, line 89: Please define a enum as follows 

```kotlin
private enum class Lookup {
    BySpecification,
    CommonDeviation
}
```
Then replace the boolean parameter

src/main/kotlin/eu/europa/ec/eudi/openid4vci/internal/DefaultAuthorizationServerMetadataResolver.kt
```diff
@@ -85,3 +78,23 @@ internal class DefaultAuthorizationServerMetadataResolver(
...
+    with(Url(this.value.toString())) {
```
babisRoutis, line 90: I can hardly follow the code and interpret `this` 
Consider using full syntax for `this` or even better use local val(s).

Also I don't get the use of the first with()

For example, looks more clean IMO

```kotlin
         URLBuilder(this).apply {
            when (lookup){
                Lookup.BySpecification ->
                    path("/${metadata.removePrefix("/").removeSuffix("/")}${pathSegments.joinToString("/")}")
                Lookup.CommonDeviation ->
                    appendPathSegments("/${metadata.removePrefix("/").removeSuffix("/")}", encodeSlash = false)
            }
        }.build()
```

src/main/kotlin/eu/europa/ec/eudi/openid4vci/internal/DefaultAuthorizationServerMetadataResolver.kt
babisRoutis: Also remove the default lookup value. Let's be explicit

babisRoutis (changes requested):

babisRoutis (approved):

PR #231, opened 2024-05-27
Deprecate ProofSigner & BindingKey

### LINKED ISSUE #232: Deprecate ProofSigner
The `ProofSigner` interface is not generic since it aims for Proof of types JWT.
It cannot be used - if needed - to model CWT proofs.
For this reason a new interface is needed.

In addition neither `BindingKey`  is generic. It represents the ways that a Pub Key can be included in a JWT Proof

babisRoutis:
This PR 
* Deprecates the name `BindingKey` favoring the more correct `JwtBindingKey`
* Deprecates the `ProofSigner` 
* Introduces a new sealed hierarchy named `PopSigner` with a single member `PopSigner.Jwt` 

The usages of `ProofSigner` have been replaced by `PopSigner` or `PopSigner.Jwt` depending on the case.

Closes #232

### DIFF
--- a/src/main/kotlin/eu/europa/ec/eudi/openid4vci/Issuance.kt
+++ b/src/main/kotlin/eu/europa/ec/eudi/openid4vci/Issuance.kt
@@ -252,7 +252,7 @@ interface RequestIssuance {
     @Deprecated(
         message = "Deprecated. Will be removed in a future release.",
         replaceWith = ReplaceWith(
-            "requestSingle(requestPayload, PopSigner.Jwt(proofSigner.getAlgorithm(),proofSigner.getBindingKey(),proofSigner)",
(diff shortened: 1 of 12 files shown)


src/main/kotlin/eu/europa/ec/eudi/openid4vci/Issuance.kt
```diff
@@ -248,6 +249,21 @@ interface RequestIssuance {
         requestPayload: IssuanceRequestPayload,
     ): Result<SubmittedRequest>
 
+    @Deprecated(
```
vafeini, line 252: Since we introduce "breaking" changes in config and batch credential request lets follow the same for the single request issuance.

PR #238, opened 2024-06-03
Support CWT Proofs

### LINKED ISSUE #237: Support CWT proofs
Currently, VCI libs support presenting JWT proofs in case an issuer requires it.

There are cases where issuer will require to provide proofs in CWT.

### LINKED ISSUE #239: Provide an option to explictly define PAR usage
Currently, library uses the PAR endpoint, during the authorization phase of authorization code flow, in case the authorization server used by the credential issuer advertises a PAR endpoint. Otherwise, library falls back to preparing an ordinary authorization request (front channel).

Ideally,  there should be a policy (or configuration option) with regards in the  usage of PAR endpoint 

There are the following option:

- Use PAR endpoint if it is being advertised, otherwise fallback to ordinary request.
- Do not use PAR endpoint. Always place ordinary front channel requests.
- Always use PAR and fail in case the authorization server doesn't provide a PAR endpoint

### LINKED ISSUE #240: Remove hard-coded `openid` scope from authorization request
While running the authorization code flow (PAR or simple), library adds to the scopes required for authorizing issuance the `openid` scope.

This is not needed and should be removed.

### LINKED ISSUE #241: Add prompt login to authorization request
During authorization code flow (PAR or simple), add `prompt=login` in order to instruct the authorization server to ask user to login to authorization issuance, regardless if his/her current session

babisRoutis:
This PR introduces support for CWT proofs. 

In particular, issuer metada for each supported configuration have been updated for CWT proofs to follow the LSP Potential Interoperability Specs like the example bellow.

```json
"proof_types_supported": {
    "cwt": {
       "proof_alg_values_supported": [ -7 ],
       "proof_crv_values_supported": [ 1 ]
     }
}
```
Note that this is a deviation from VCI Draft 13 where in paragraph [7.2.1.3](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#name-cwt-proof-type) requires that the metadata for CWT proofs should be

> Cryptographic algorithm names used in the proof_signing_alg_values_supported Credential Issuer metadata parameter for this proof type SHOULD be one of those defined in [[IANA.COSE.ALGS](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#IANA.COSE.ALGS)].

Furthermore the PR:
* Introduces `PopSigner.Cwt` to the public API to allow a caller to sign the CWT 
* Introduces a CwtProofBuilder (internal class) to form the CWT and delegate signing to caller-provided `PopSigner.Cwt`.
* Introduces a configuration option to define the policy of using PAR endpoint

Closes #237 
Closes #239 
Closes #240
Closes #241

### DIFF
--- a/src/main/kotlin/eu/europa/ec/eudi/openid4vci/Config.kt
+++ b/src/main/kotlin/eu/europa/ec/eudi/openid4vci/Config.kt
@@ -33,6 +33,8 @@ typealias ClientId = String
  * by the credential issuer and [AuthorizeIssuanceConfig.FAVOR_SCOPES] is selected then scopes will be used.
  * Otherwise, authorization details (RAR)
  * @param dPoPSigner a signer that if provided will enable the use of DPoP JWT
+ * @param parUsage whether to use PAR in case of authorization code grant
+ * @param clock Wallet's clock
  */
 data class OpenId4VCIConfig(
     val clientId: ClientId,
@@ -41,6 +43,7 @@ data class OpenId4VCIConfig(
     val credentialResponseEncryptionPolicy: CredentialResponseEncryptionPolicy,
     val authorizeIssuanceConfig: AuthorizeIssuanceConfig = AuthorizeIssuanceConfig.FAVOR_SCOPES,
     val dPoPSigner: PopSigner.Jwt? = null,
+    val parUsage: ParUsage = ParUsage.IfSupported,
     val clock: Clock = Clock.systemDefaultZone(),
 ) {
 
@@ -55,6 +58,20 @@ data class OpenId4VCIConfig(
             }
         }
     }
+}
+
+/**
+ * Wallet's policy in regard to using PAR, during a authorization code grant.
+ * - [IfSupported]: If authorization server advertises PAR endpoint it will be used. Otherwise, falls back
+ *   to usual authorization code flow
+ * - [DonNot]: Disables PAR. Wallet will use the usual authorization code flow
+ * - [Always]: Wallet always will place PAR request, regardless what if authorization server advertises the PAR
+ *   endpoint
+ */
+enum class ParUsage {
+    IfSupported,
+    DonNot,
+    Always,
 }
 
 /**
--- a/src/main/kotlin/eu/europa/ec/eudi/openid4vci/Config.kt
+++ b/src/main/kotlin/eu/europa/ec/eudi/openid4vci/Config.kt
@@ -65,13 +65,13 @@ data class OpenId4VCIConfig(
  * - [IfSupported]: If authorization server advertises PAR endpoint it will be used. Otherwise, falls back
  *   to usual authorization code flow
  * - [Never]: Disables PAR. Wallet will use the usual authorization code flow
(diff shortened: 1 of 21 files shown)


src/main/kotlin/eu/europa/ec/eudi/openid4vci/internal/http/AuthorizationServerClient.kt
```diff
@@ -178,11 +186,12 @@ internal class AuthorizationServerClient(
                 state(State(state))
                 issuerState?.let { customParameter("issuer_state", issuerState) }
                 if (scopes.isNotEmpty()) {
-                    scope(NimbusScope(*scopes.map { it.value }.toTypedArray() + "openid"))
+                    scope(NimbusScope(*scopes.map { it.value }.toTypedArray()))
                 }
                 if (credentialsConfigurationIds.isNotEmpty()) {
                     authorizationDetails(credentialsConfigurationIds.map(::toNimbus))
                 }
+                prompt(Prompt.Type.LOGIN)
```
dzarras, line 194: Something minor: Given that the Authorization Server might not advertise PAR support, and `ParUage` might be set to `Always`, we could add a check that `parEndpoint` is indeed not null.

```kotlin
val parEndpoint = requireNotNull(authorizationServerMetadata.pushedAuthorizationRequestEndpointURI)
```

Even without this, at the end of the day the constructor call to `PushedAuthorizationRequest` will fail non the less.

src/main/kotlin/eu/europa/ec/eudi/openid4vci/Config.kt
```diff
@@ -57,6 +60,20 @@ data class OpenId4VCIConfig(
...
+ * - [IfSupported]: If authorization server advertises PAR endpoint it will be used. Otherwise, falls back
+ *   to usual authorization code flow
+ * - [Never]: Disables PAR. Wallet will use the usual authorization code flow
+ * - [Always]: Wallet always will place PAR request, regardless what if authorization server advertises the PAR
+ *   endpoint
+ */
+enum class ParUsage {
+    IfSupported,
+    Never,
+    Always,
```
dzarras, line 74: Perhaps `Always` could be renamed to `Required` instead?

src/main/kotlin/eu/europa/ec/eudi/openid4vci/internal/http/AuthorizationServerClient.kt
babisRoutis: So, perhaps this means that "Always" doesn't make sense.

babisRoutis: @dzarras Please check

src/main/kotlin/eu/europa/ec/eudi/openid4vci/Config.kt
babisRoutis: Done

dzarras (dismissed):

PR #281, opened 2024-07-07
Support attestation-based client authentication

### LINKED ISSUE #304: Support Attestation-Based Client Authentication
HAIP [requires](https://openid.github.io/oid4vc-haip-sd-jwt-vc/openid4vc-high-assurance-interoperability-profile-sd-jwt-vc-wg-draft.html#name-token-endpoint) the use of [OAuth 2.0 Attestation-Based Client Authentication](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-attestation-based-client-auth-03).

Library should be able to support this kind of client/wallet authentication.

In scope of this issue is the ability to use a [Client Attestation JWT](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-attestation-based-client-auth-03#name-client-attestation-jwt) (issued by wallet provider)   which basically means 
- be able to create [Client Attestation PoP JWT](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-attestation-based-client-auth-03#name-client-attestation-pop-jwt)
- Populate the relevant [HTTP headers](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-attestation-based-client-auth-03#name-client-attestation-http-hea) when accessing PAR and/or Token Endpoint

Out of scope, 
- provision of Client Attestation JWT. Library assumes that wallet has already obtained a Client Attestation

babisRoutis:
PR aims (hopefully) to support attestation-base client (wallet) authentication.

- [x] Introduce a sealed hierarchy named `Client` having two member `Public` and `Attested`. `Client` would express the OAUTH2 client and in particular the way that it can be authenticated to token and/or PAR endpoint
- [x] Assumed that `Client.Attested` has already a `ClientAttestationJWT`. How this is provisioned is outside the scope of the PR
- [x] Change the `TokenEndPointClient` to create a `ClientAttestation` by producing the POP JWT 
- [x] Similarly, change the PAR request  
- [x] Replace `clientId` with `client` in `OpenId4VCIConfig` 
- [x] Take into account [Wallet Attestation Schema](https://openid.github.io/oid4vc-haip-sd-jwt-vc/openid4vc-high-assurance-interoperability-profile-sd-jwt-vc-wg-draft.html#name-wallet-attestation-schema) from HAIP
- [x] Fix the DeferredIssuerExt.kt extensions

Closes #304


babisRoutis:
Attestation based client authentication [states](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-attestation-based-client-auth-03#name-reuse-of-a-client-attestati) that

> Implementers should be aware that the design of this authentication mechanism deliberately allows for a Client Instance to re-use a single Client Attestation JWT in multiple interactions/requests with an Authorization Server, **whilst producing a fresh Client Attestation PoP JWT**. Client deployments should consider this when determining the validity period for issued Client Attestation JWTs as this ultimately controls how long a Client Instance can re-use a single Client Attestation JWT.

To my understanding, the highlighted phrase means that library should produce a new PoP JWT with each interaction.
For instance, in a typical HAIP scenario 
- Wallet will place a PAR, then
- Will get from token endpoint the `access_token`

Those two interactions, will use different PoP JWT, yet bound to the same client attestation JWT.

PR #287, opened 2024-07-12
Remove deprecated code

### LINKED ISSUE #276: Remove features and code deprecated in v0.3.x
Features that have been deprecated in v0.3.x include:

- [x] `SubmittedRequest` typealias
- [x] Remove deprecated abstraction for `ProofSigner`
- [x] Remove deprecated methods from `RequestIssuance`, enforcing usage of the new methods that behave more concisely
- [x] Remove deprecated methods from `RequestBatchIssuance`, enforcing usage of the new methods that behave more concisely
- [x] Remove `SubmissionOutcome.InvalidProof` and report `invalid_proof` error as a failure. After the introduction of the new methods for `RequestIssuance` and `RequestBatchIssuance` which retry automatically in case of `invalid_proof` the `final invalid_proof` is probably is inrecoverable

### LINKED ISSUE #288: Differantiate `invalid_proof` error
After the introduction to the library of the ability to retry on `invalid_proof` error, provided that caller has given proofs, 
it is possible to further refine this error as follows:

`CredentialIssuanceError.InvalidProof` 

It is raised to :
- Indicate that c_nonce is required for a credential request, yet the token endpoint of the credential issuer didn't provide this 
- Caller didn't provide  proof when placing the request (parameter is optional), thus library cannot retry automatically the call 

`CredentialIssuanceError.IrrecoverableInvalidProof`
- Indicates that library was able to retry a credential request and got again an `invalid_proof`. Effectively caller cannot recover from this error.

babisRoutis:
Closes #276 
Closes #288

The most significant change is the removal of `SubmissionOutcome.InvalidProof`

After the introduction of the library feature for automatically retrying (if possible) in case of an `invalid_proof` error, this error is being reported to the caller as `SubmissionOutcome.Failed` having an error of type `CredentialIssuanceError.InvalidProof` 

If caller has provided the proofs, an `InvalidProof` means that the credential issuer indeed rejected proof because actually evaluates them as erroneous. This special case is reported with `IrrecoverableInvalidProof`

### DIFF
--- a/src/main/kotlin/eu/europa/ec/eudi/openid4vci/Issuance.kt
+++ b/src/main/kotlin/eu/europa/ec/eudi/openid4vci/Issuance.kt
@@ -123,6 +123,15 @@ typealias AuthorizedRequestAnd<T> = Pair<AuthorizedRequest, T>
  */
 interface RequestIssuance {
 
+    @Deprecated(
+        message = "Deprecated and will be removed",
+        replaceWith = ReplaceWith("requestSingle(requestPayload, popSigner)"),
+    )
+    suspend fun AuthorizedRequest.requestSingleAndUpdateState(
+        requestPayload: IssuanceRequestPayload,
+        popSigner: PopSigner?,
+    ): Result<AuthorizedRequestAnd<SubmissionOutcome>> = requestSingle(requestPayload, popSigner)
+
     /**
      * Places a request to the credential issuance endpoint.
      * Method will attempt to automatically retry submission in case
@@ -139,7 +148,7 @@ interface RequestIssuance {
      * @return the possibly updated [AuthorizedRequest] (if updated it will contain a fresh c_nonce) and
      * the [SubmissionOutcome]
      */
-    suspend fun AuthorizedRequest.requestSingleAndUpdateState(
+    suspend fun AuthorizedRequest.requestSingle(
         requestPayload: IssuanceRequestPayload,
         popSigner: PopSigner?,
     ): Result<AuthorizedRequestAnd<SubmissionOutcome>>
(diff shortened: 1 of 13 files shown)


src/main/kotlin/eu/europa/ec/eudi/openid4vci/Issuance.kt
```diff
@@ -235,40 +156,6 @@ interface RequestBatchIssuance {
     suspend fun AuthorizedRequest.requestBatchAndUpdateState(
```
vafeini, line 156: Shall this be renamed to AuthorizedRequest.requestBatch()? Since the default behavior is to update the state internally I guess there is no need to state it on the method name

src/main/kotlin/eu/europa/ec/eudi/openid4vci/Issuance.kt
```diff
@@ -165,7 +134,7 @@ interface RequestIssuance {
      * @param popSigner Signer component of the proof to be sent. Although this is an optional
      * parameter, only required in case the present authorization state is [AuthorizedRequest.ProofRequired],
      * caller is advised to provide it, in order to allow the method to automatically retry
-     * in case of [SubmissionOutcome.InvalidProof]
+     * in case of [CredentialIssuanceError.InvalidProof]
      *
      * @return the possibly updated [AuthorizedRequest] (if updated it will contain a fresh c_nonce) and
      * the [SubmissionOutcome]
```
vafeini, line 140: Shall this be renamed to AuthorizedRequest.requestBatch()? Since the default behavior is to update the state internally I guess there is no need to state it on the method name

vafeini:
Maybe not in the context of this PR but shall we consider making AuthorizedRequest.ProofRequired and AuthorizedRequest.NoProofRequired internal states not exposed to public.

vafeini (approved):

babisRoutis:
> Maybe not in the context of this PR but shall we consider making AuthorizedRequest.ProofRequired and AuthorizedRequest.NoProofRequired internal states not exposed to public.

Let's leave this outside the scope of the PR

There are open discussion in OpenId4VCI spec with regards to the use of `c_nonce`. In a future release proofs may not require it.

PR #332, opened 2024-10-25
Add support for authorization server-provided `dpopNonce`

### LINKED ISSUE #330: Support Nonce in dPoP request header in token endpoint
**Reference:** https://datatracker.ietf.org/doc/html/rfc9449

In `TokenEndpointClient.kt`,  `fun requestAccessToken(..)`: 

```
            httpClient.submitForm(tokenEndpoint.toString(), formParameters) {
                dPoPJwtFactory?.let { factory ->
                    dpop(factory, tokenEndpoint, Htm.POST, accessToken = null, nonce = null)
                }
```
The nonce is set to null. 

Can we allow `fun AuthorizationRequestPrepared.authorizeWithAuthorizationCode()` in `AuthorizeIssuance.kt` to accept a nonce as nullable string parameter and then allow `fun requestAccessToken()` to take the same parameter to be set in the dPoP request header?

Can I raise a PR with the above changes that you can review?

babisRoutis:
This PR adds support for using `dpopNonce` provided by the authorization server, in the following interactions:
- When requesting access token (in the context of authorization code grant and / or pre-authorized code grant)
- When refreshing an access token (in the context of refresh grant).
- When pushing an authorization request to PAR endpoint

In the above cases, authorization server may reply with a (oaut2) error  `use_dpop_nonce` either because the `dopNonce` is missing or because the the value it is not the correct.

Library catches the above error, updates the `dpopNonce` of the client, to the new value provided by the server and repeats the call using the new `dpopNonce`. At this stage, if authorization server replies again with a `use_dpop_nonce`, library will not retry again.

Furthermore, authorization server at any of the aforementioned three interactions may choose to update the `dpopNonce`.
To satisfy this, library evaluates whether success (`200`) responses contain a `dpopNonce` header, and if present, the locally held `dpopNonce` is updated.

PS: On the client side, the `dpopNonce` is kept within the `AuthorizedRequest`. Since the library is stateless, all the above interactions require from the caller to provide an `AuthorizedRequest` and return a possibly updated `AuthorizedRequest`

Closes #330

### DIFF
--- a/src/main/kotlin/eu/europa/ec/eudi/openid4vci/AuthorizeIssuance.kt
+++ b/src/main/kotlin/eu/europa/ec/eudi/openid4vci/AuthorizeIssuance.kt
@@ -15,6 +15,7 @@
  */
 package eu.europa.ec.eudi.openid4vci
 
+import com.nimbusds.openid.connect.sdk.Nonce
 import java.time.Instant
 
 /**
@@ -34,6 +35,7 @@ data class AuthorizationRequestPrepared(
     val pkceVerifier: PKCEVerifier,
     val state: String,
     val identifiersSentAsAuthDetails: List<CredentialConfigurationIdentifier>,
+    val dpopNonce: Nonce?,
 ) : java.io.Serializable
 
 /**
@@ -49,6 +51,7 @@ sealed interface AuthorizedRequest : java.io.Serializable {
     val refreshToken: RefreshToken?
     val credentialIdentifiers: Map<CredentialConfigurationIdentifier, List<CredentialIdentifier>>?
     val timestamp: Instant
+    val dpopNonce: Nonce?
 
     fun isAccessTokenExpired(at: Instant): Boolean = accessToken.isExpired(timestamp, at)
     fun isRefreshTokenExpiredOrMissing(at: Instant): Boolean = refreshToken?.isExpired(timestamp, at) ?: true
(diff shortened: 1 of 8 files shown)


dzarras:
@babisRoutis Looks good to me. 

Could you please though document the new DPoP Nonce parameter wherever needed in order to be aligned with the remainder of the documentation?

*Thinking out loudly:* Does it makes senses to also include the DPoP Nonce in `DeferredIssuanceStoredContextTO`? Currently we don't. Most probably it doesn't make sense to do so, since by the time we might have to reach out to the Authorization Server (in case of an Access Token refresh), the provided DPoP Nonce will most probably be expired.

babisRoutis:
> @babisRoutis Looks good to me.
> 
> Could you please though document the new DPoP Nonce parameter wherever needed in order to be aligned with the remainder of the documentation?
> 
> _Thinking out loudly:_ Does it makes senses to also include the DPoP Nonce in `DeferredIssuanceStoredContextTO`? Currently we don't. Most probably it doesn't make sense to do so, since by the time we might have to reach out to the Authorization Server (in case of an Access Token refresh), the provided DPoP Nonce will most probably be expired.

@dzarras Added a description on the PR.

babisRoutis:
> _Thinking out loudly:_ Does it makes senses to also include the DPoP Nonce in `DeferredIssuanceStoredContextTO`? Currently we don't. Most probably it doesn't make sense to do so, since by the time we might have to reach out to the Authorization Server (in case of an Access Token refresh), the provided DPoP Nonce will most probably be expired.

We don't need to keep `dpopNonce` in the case of deferred issuance. If required by the authorization server, library already can recover the `use_dpop_nonce` error and repeat the call

dzarras (approved):

babisRoutis:
@dzarras Can you please check https://ktor.io/docs/client-http-send.html ?
It seems to me that we can create an interceptor to retry on time with HttpSend ktor plugin. I think retry times is configurable

babisRoutis:
> @dzarras Can you please check https://ktor.io/docs/client-http-send.html ? It seems to me that we can create an interceptor to retry on time with HttpSend ktor plugin. I think retry times is configurable

Perhaps we can leave this outside the PR scope

dzarras:
> @dzarras Can you please check https://ktor.io/docs/client-http-send.html ? It seems to me that we can create an interceptor to retry on time with HttpSend ktor plugin. I think retry times is configurable

The only configuration option I found for this plugin is: https://api.ktor.io/ktor-client/ktor-client-core/io.ktor.client.plugins/-http-send/-config/index.html `maxSendCount` which doesn't seem to be the retry count.

PR #335, opened 2024-10-30
DPoP: Resource Server-Provided Nonce

### LINKED ISSUE #334: Add support for Resource Server-Provided Nonce
Affects:
- Credential endpoint
- Deferred endpoint
- Notification endpoint
- c_nonce endpoint (introduced with latest OpenId4VCI)

Reference: [Resource Server-Provided Nonce](https://datatracker.ietf.org/doc/html/rfc9449#name-resource-server-provided-no)

dzarras:
Adds support for Resource-Server provided Nonce in DPoP to the following:

- Credential Endpoint
- Deferred Credential Endpoint
- Notification Endpoint

Follows a similar approach to #332.

Closes #334


dzarras:
Hi @daniel-idrees,

Could you please also have a look at this PR and let us know of any comments?

Thanks.

src/main/kotlin/eu/europa/ec/eudi/openid4vci/DPoP.kt
```diff
@@ -143,14 +143,15 @@ fun HttpRequestBuilder.bearerOrDPoPAuth(
     htu: URL,
     htm: Htm,
     accessToken: AccessToken,
+    nonce: String?,
```
babisRoutis, line 146: I think that we need to keep to a single representation for (DPoP) Nonce
Options are:
1. Use a simple String
2. Typealias Nonce = String
3. Value class Nonce
4. Use of Nonce class from Nimbus

In the first PR, we used option 4, but probably it is not ideal to couple our public API to an external class.
I would go with the value class approach

src/main/kotlin/eu/europa/ec/eudi/openid4vci/internal/RequestIssuanceImpl.kt
```diff
@@ -151,10 +152,11 @@ internal class RequestIssuanceImpl(
 
     private suspend fun placeIssuanceRequest(
         token: AccessToken,
+        dpopNonce: Nonce?,
```
babisRoutis, line 155: Please consider renaming this to `resourceServerDpopNonce`

src/main/kotlin/eu/europa/ec/eudi/openid4vci/NotifyIssuer.kt
```diff
@@ -63,6 +63,6 @@ fun interface NotifyIssuer {
          * @return a [NotifyIssuer]
          */
         internal operator fun invoke(notificationEndPointClient: NotificationEndPointClient): NotifyIssuer =
-            NotifyIssuer { event -> notificationEndPointClient.notifyIssuer(accessToken, event) }
+            NotifyIssuer { event -> notificationEndPointClient.notifyIssuer(accessToken, resourceServerDpopNonce, event) }
```
babisRoutis, line 66: The `notify` methods has to be changed, I think.
Instead of returning a `Result<Unit>`,  it should return the possibly updated `AuthorizedRequest`. Perhaps `AuthorizedRequestAnd<Unit>`

babisRoutis (changes requested):

src/main/kotlin/eu/europa/ec/eudi/openid4vci/DPoP.kt
dzarras: Introduce a new `Nonce` value class.

babisRoutis (approved):

babisRoutis:
@dzarras Approved this.

Just two minor comments
- On second thought returning `Result<AuthorizedRequestAnd<Unit>>` seems somehow awkward. If you agree, then just turn a `Result<AuthorizedRequest>`
- We can also update the README, in particular the section for DPoP to highlight 
  - Support to both Authorization & Resource Server Provided DPoP Nonce
  - The feature of the library to refresh DPoP Nonce on `200`
  - The feature of the library to recover `400` (error `use_dpop_nonce`) responses while interacting with authorization server
  - The feature of the library to recover `401` (error `use_dpop_nonce`) responses while interacting with credential issuer

daniel-idrees:
@dzarras Great Work! Thank you for a quick implementation. I was able to test the credential endpoint and it worked for us with the new dpopnonce handling.

PR #339, opened 2024-11-11
Conditionally add iss claim to proof

### LINKED ISSUE #338: iss claim in jwt proof type must be omitted for pre-authorized code flow according to the openid4vci spec
iss: OPTIONAL (string). The value of this claim MUST be the client_id of the Client making the Credential request. This claim MUST be omitted if the access token authorizing the issuance call was obtained from a Pre-Authorized Code Flow through anonymous access to the token endpoint.

https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0-14.html#section-7.2.1.1-2.2.1

babisRoutis:
Closes #338

### DIFF
--- a/src/main/kotlin/eu/europa/ec/eudi/openid4vci/AuthorizeIssuance.kt
+++ b/src/main/kotlin/eu/europa/ec/eudi/openid4vci/AuthorizeIssuance.kt
@@ -15,8 +15,6 @@
  */
 package eu.europa.ec.eudi.openid4vci
 
-import kotlinx.serialization.SerialName
-import kotlinx.serialization.Serializable
 import java.time.Instant
 
 /**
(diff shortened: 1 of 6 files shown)


src/main/kotlin/eu/europa/ec/eudi/openid4vci/AuthorizeIssuance.kt
```diff
@@ -37,6 +39,15 @@ data class AuthorizationRequestPrepared(
     val dpopNonce: Nonce?,
 ) : java.io.Serializable
 
+@Serializable
```
dzarras, line 42: @babisRoutis I don't believe we need to mark this as JSON Serializable. Isn't this enum part of the core domain? 
Also, even though I think it's not needed for enum, please make this enum extend java.io.Serializable.

dzarras: Just noticed we use this enum is `DeferredIssuanceStoredContextTO`. Would it make sense to introduce a separate enum for serialization, and avoid Kotlin Serialization annotations in the core domain?

babisRoutis: They idea was to use it with Deferred Issuer.
Anyway, implemented `java.io.Serializable` removed kotlinx serializable.

dzarras (approved):

PR #378, opened 2025-02-12
Ensure Authorization Server support Authorization Code Flow when Credential Offer requires it.

### LINKED ISSUE #373: NPE when authorization_endpoint is missing
If Authorization Server Metadata does not contain `authorization_endpoint`, AuthorizationEndpointClient will throw a null pointer exception on line `authorizationServerMetadata.authorizationEndpointURI.toURL()`.

dzarras:
Closes #373

### DIFF
--- a/src/main/kotlin/eu/europa/ec/eudi/openid4vci/CredentialOfferRequestResolver.kt
+++ b/src/main/kotlin/eu/europa/ec/eudi/openid4vci/CredentialOfferRequestResolver.kt
@@ -32,7 +32,7 @@ data class CredentialOffer(
 ) : Serializable {
     init {
         require(credentialConfigurationIdentifiers.isNotEmpty()) { "credentials must not be empty" }
-        if (null == grants || grants is Grants.AuthorizationCode) {
+        if (grants is Grants.AuthorizationCode) {
             requireNotNull(authorizationServerMetadata.authorizationEndpointURI) {
                 "Credential Offer requires Authorization Code Grant, but the Authorization Server does not support it"
             }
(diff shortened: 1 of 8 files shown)


src/main/kotlin/eu/europa/ec/eudi/openid4vci/CredentialOfferRequestResolver.kt
```diff
@@ -32,6 +32,11 @@ data class CredentialOffer(
 ) : Serializable {
     init {
         require(credentialConfigurationIdentifiers.isNotEmpty()) { "credentials must not be empty" }
+        if (null == grants || grants is Grants.AuthorizationCode) {
```
babisRoutis, line 35: Why the `null == grants` check?

src/main/kotlin/eu/europa/ec/eudi/openid4vci/internal/DefaultCredentialOfferRequestResolver.kt
```diff
@@ -103,6 +103,15 @@ internal class DefaultCredentialOfferRequestResolver(
         val grants = credentialOffer.grants?.toGrants(credentialIssuerMetadata)
         val authorizationServer = grants?.authServer() ?: credentialIssuerMetadata.authorizationServers[0]
         val authorizationServerMetadata = fetchAuthServerMetaData(authorizationServer)
+        if (null == grants || grants is Grants.AuthorizationCode) {
```
babisRoutis, line 106: Why the `null == grants` check?

babisRoutis (dismissed):
@dzarras I am not sure about the `grants == null` check that appears in couple places.

The spec states the following:

> If grants is not present or is empty, the Wallet MUST determine the Grant Types the Credential Issuer's Authorization Server supports using the respective metadata


Currently, we don't support this functionality in our library, and certainly the absence of `grants` cannot be translated as a value that default to the authorization code grant.

Issue #375, opened 2025-02-12
Update Credential Issuer Metadata


vafeini:
Credential Issuer metadata parsing must be updated as follows

- [x] Introduce `nonce_endpoint` metadata attribute to advertice nonce endpoint url (will be done in the context of #372)
- [x] Update credential configurations (element 'credential_configurations_supported'):
   1. `claims` of credential configurations (all format profiles) must be expresed with claim path pointers according to section [B.2](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0-15.html#appendix-B.2)
   2. `display` object include `background_image` attibute (see [here](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0-15.html#section-11.2.3-2.11.2.6.2.6.1))
   3. Remove `order` attribute for format specific configurations
   4. Update definition of `credential_definition` in W3C related profiles 
   5. Include in 'proof_types_supported' object a new attribute named 'key_attestations_required' as described [here](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0-15.html#section-11.2.3-2.11.2.5.2.2.1)

dzarras:
Please also note that `claim` is no longer a JSON Object, but rather an JSON Array.
It's also defined per profile, meaning that `claims` is not a property mandated by OpenId4VCI for all profiles.

References: 
* [credential_configurations_supported](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0-15.html#section-11.2.3-2.11.1)
* [MSO MDoc `claims`](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0-15.html#appendix-A.2.2-3.2)
* [SD-JWT VC `claims`](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0-15.html#appendix-A.3.2-3.2)

PR #382, opened 2025-02-19
Update credential issuer metadata to draft 15

vafeini:
Tackles #375

### DIFF
--- a/src/main/kotlin/eu/europa/ec/eudi/openid4vci/CredentialConfiguration.kt
+++ b/src/main/kotlin/eu/europa/ec/eudi/openid4vci/CredentialConfiguration.kt
@@ -75,8 +75,8 @@ sealed interface ProofTypeMeta : Serializable {
 
 sealed interface KeyAttestationRequirement {
 
-    data object NoRequirement : KeyAttestationRequirement {
-        private fun readResolve(): Any = NoRequirement
+    data object NotRequired : KeyAttestationRequirement {
+        private fun readResolve(): Any = NotRequired
     }
 
     data object RequiredNoConstraints : KeyAttestationRequirement {
--- a/src/main/kotlin/eu/europa/ec/eudi/openid4vci/CredentialConfiguration.kt
+++ b/src/main/kotlin/eu/europa/ec/eudi/openid4vci/CredentialConfiguration.kt
@@ -57,7 +57,10 @@ enum class ProofType : Serializable {
 }
 
 sealed interface ProofTypeMeta : Serializable {
-    data class Jwt(val algorithms: List<JWSAlgorithm>) : ProofTypeMeta {
+    data class Jwt(
+        val algorithms: List<JWSAlgorithm>,
+        val keyAttestationRequirement: KeyAttestationRequirement,
+    ) : ProofTypeMeta {
         init {
             require(algorithms.isNotEmpty()) { "Supported algorithms in case of JWT cannot be empty" }
         }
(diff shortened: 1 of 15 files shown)


src/main/kotlin/eu/europa/ec/eudi/openid4vci/CredentialConfiguration.kt
```diff
@@ -124,15 +150,16 @@ sealed interface CredentialConfiguration : Serializable {
     val credentialSigningAlgorithmsSupported: List<String>
     val proofTypesSupported: ProofTypesSupported
     val display: List<Display>
+    val claims: List<Claim>?
```
dzarras, line 153: @vafeini I'm not sure if `claims` should be defined here. It's just happens that all the Credential Format Profiles mentioned in the spec, define the additional `credential_configurations_supported` property `claims`.

References:
* https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#section-11.2.3-2.11.1
* https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#appendix-A.1.1.2-3.1
* https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#appendix-A.2.2-3.2
* https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#appendix-A.3.2-3.2

babisRoutis: @dzarras 

It is true that `claims` is not defined as a common (among formats) attribute, simply because it is not included in https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#section-11.2.3-2.11.1.

On the other hand, there was an intentional effort to keep the same definition of `claims` across the formats.
I would recommend keeping it as it is

dzarras (approved):

src/main/kotlin/eu/europa/ec/eudi/openid4vci/CredentialConfiguration.kt
```diff
@@ -70,6 +73,28 @@ sealed interface ProofTypeMeta : Serializable {
     data class Unsupported(val type: String) : ProofTypeMeta
 }
 
+sealed interface KeyAttestationRequirement {
+
+    data object NoRequirement : KeyAttestationRequirement {
```
babisRoutis, line 78: `NoRequirement` or `NotRequired` ?

src/main/kotlin/eu/europa/ec/eudi/openid4vci/internal/http/CredentialIssuerMetadataJsonParser.kt
```diff
@@ -56,15 +53,22 @@ private sealed interface CredentialSupportedTO {
...
+    val proofTypesSupported: Map<String, ProofTypeSupportedMetaTO>?
     val display: List<CredentialSupportedDisplayTO>?
 
     fun toDomain(): CredentialConfiguration
 }
 
 @Serializable
-private data class ProofSigningAlgorithmsSupportedTO(
+private data class ProofTypeSupportedMetaTO(
     @SerialName("proof_signing_alg_values_supported") val algorithms: List<String> = emptyList(),
```
babisRoutis, line 64: Is it empty list or null?

babisRoutis (changes requested):

src/main/kotlin/eu/europa/ec/eudi/openid4vci/CredentialConfiguration.kt
vafeini: NotRequired is better, I will change it

src/main/kotlin/eu/europa/ec/eudi/openid4vci/internal/http/CredentialIssuerMetadataJsonParser.kt
vafeini: `proof_signing_alg_values_supported` is a mandatory attribute if `proof_types_supported` are specified in credential configuration. So it cannot be null. Issuer must specify it even if the array is empty. An empty array has no meaning so we do validate it when converting TOs to domain objects.

PR #419, opened 2025-06-18
Signers

### LINKED ISSUE #421: Change how the wallet defer sign operations
The wallet may use keys that are managed by Android Key Store or another Cloud base HSM.

VCI lib currently excepts a `List<PoPSigner>` to be provided by the wallet in order to create the Proofs of type JWT, during the issuance a credential configuration.

There several problems to the List<PoPSigner>
1) It is JWS specific, given that it is implemented as a JWSSigner interface from Nimbus
2) It doesn't support a single authentication for a batch of signing operations

To address the above problems we need to introduce some new abstractions:

`SignOperation` represents a defer signing for a specific managed key (regardless of the managing environment : KeyStore or Cloud HSM)

```kotlin
fun interface SignOperation {
    suspend fun sign(input: ByteArray): Result<ByteArray>
}
```

`SignOp` wraps a sign operation together with the public material of the respective key.
It is parameterized to the representation of the public material. It can be a public key (represented as a JCA class), or a JWK, or a certificate chain (to support the attested keys)

```kotlin
data class SignOp<out PUB>(
    val signingAlgorithm: String,
    val operation: SignOperation,
    val publicMaterial: PUB,
)
```

`Signer` is what VCI lib should be provided for a single signing operation. 
The important point is that `authenticate` function is a placeholder to defer the provisioning of the signing operation (which is also deferred). This can be used by a wallet to authorize the usage (or operation) of a key, if needed.

```kotlin
fun interface Signer<out PUB> {
    suspend fun authenticate(): SignOp<PUB>
}
```

This is the generalization of `SignOp` in case we are working with a set (batch) of keys that should be used together (that the JWT proofs case)

```kotlin
data class BatchSignOp<out PUB>(
    val operations: List<SignOp<PUB>>,
)
```

Finally, 

`BatchSigner` is the generalization of `Signer` in the case of a batch of keys.

Note that BatchSigner is not a `List<Signer>`. Such a model implies that each key must be authorized explictily.

Rather, `BatchSigner` acquires the list of signing operation after a SINGLE authorization event.

```kotlin
interface BatchSigner<out PUB> {
    suspend fun authenticate(): BatchSignOp<PUB>
}
```

gdimtsas:
Closes #421

### DIFF
--- a/src/main/kotlin/eu/europa/ec/eudi/openid4vci/Signers.kt
+++ b/src/main/kotlin/eu/europa/ec/eudi/openid4vci/Signers.kt
@@ -57,6 +57,9 @@ data class BatchSignOperation<out PUB>(
  */
 interface Signer<out PUB> {
 
+    /**
+     * The algorithm that will be used for signing
+     */
     val javaAlgorithm: String
 
     /**
@@ -92,6 +95,9 @@ interface Signer<out PUB> {
  */
 interface BatchSigner<out PUB> {
 
+    /**
+     * The algorithm that will be used for signing
+     */
     val javaAlgorithm: String
 
     /**
--- a/src/main/kotlin/eu/europa/ec/eudi/openid4vci/internal/JwtProofSigners.kt
+++ b/src/main/kotlin/eu/europa/ec/eudi/openid4vci/internal/JwtProofSigners.kt
@@ -27,8 +27,8 @@ internal data class JwtProofClaims(
     @SerialName("aud") val audience: String,
     @Serializable(with = NumericInstantSerializer::class)
     @SerialName("iat") val issuedAt: Instant,
-    @SerialName("iss") val issuer: String?,
-    @SerialName("nonce") val nonce: String?,
+    @SerialName("iss") val issuer: String? = null,
+    @SerialName("nonce") val nonce: String? = null,
 )
 
 internal class JwtProofSigner(
--- a/src/main/kotlin/eu/europa/ec/eudi/openid4vci/internal/RequestIssuanceImpl.kt
+++ b/src/main/kotlin/eu/europa/ec/eudi/openid4vci/internal/RequestIssuanceImpl.kt
@@ -177,7 +177,7 @@ internal class RequestIssuanceImpl(
         }
     }
 
-    private fun BatchSignOperation<JwtBindingKey>.assertMatchesBatchIssuanceBatchSize() {
+    private fun BatchSignOperation<JwtBindingKey>.assertMatchesBatchIssuanceBatchSize() =
         when (val popSignersNo = operations.size) {
             0 -> error("At least one PopSigner is required in Authorized.ProofRequired")
             1 -> Unit
@@ -193,7 +193,6 @@ internal class RequestIssuanceImpl(
                 }
             }
         }
-    }
 
     private fun String.toSupportedJoseAlgorithm(credentialConfigId: CredentialConfigurationIdentifier): JWSAlgorithm {
         val proofTypesSupported = credentialSupportedById(credentialConfigId).proofTypesSupported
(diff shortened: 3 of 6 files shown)


babisRoutis:
@gdimtsas 

To be implemented:
-  Introduce a new data class Named JwtProofClaims  (serializable via kotlinx.serialization)
-  PopSigner --> NewPopSginer  A convert from existing PopSigner to the new PopSigner
-  Create  a new class Named JwtProofsSigner

```kotlin
class JwtProofSigner() {
    suspend fun sign(claims: JwtProofClaims) : List<Pair<EcJWK, JWT>>
}
```

src/main/kotlin/eu/europa/ec/eudi/openid4vci/Signers.kt
```diff
@@ -0,0 +1,304 @@
...
+                    is JwtBindingKey.Jwk -> {
+                        put("jwk", pubKey.jwk.asJsonElement())
+                    }
+                    is JwtBindingKey.X509 -> {
+                        put("x5c", pubKey.chain.asJsonElement())
+                    }
+                }
+                customizeHeader(pubKey)
+            }
+            val claimsJson = Json.encodeToJsonElement(claims).jsonObject
```
babisRoutis, line 213: This can go outside the loop. The same claim is being signed multiple tiples

vafeini (changes requested):

src/main/kotlin/eu/europa/ec/eudi/openid4vci/Signers.kt
```diff
@@ -0,0 +1,116 @@
...
+                    }
+            }
+    }
+}
+
+class DefaultJwtSigner<Claims, PUB>(
+    private val serializer: KSerializer<Claims>,
+    private val signOp: SignOp<PUB>,
+    private val customizeHeader: JsonObjectBuilder.(PUB) -> Unit = {},
+    private val assertions: SignOp<PUB>.() -> Unit = {},
```
babisRoutis, line 105: Assertions can be applied to the caller side. That's not c JwtSigner concern

src/main/kotlin/eu/europa/ec/eudi/openid4vci/Signers.kt
```diff
@@ -0,0 +1,116 @@
...
+
+fun interface JwtBatchSigner<in Claims> {
+    suspend fun sign(claims: Claims): List<String>
+
+    companion object {
+
+        inline operator fun <reified Claims, PUB> invoke(
+            signOps: BatchSignOp<PUB>,
+            noinline customizeHeader: JsonObjectBuilder.(PUB) -> Unit = {},
+            noinline assertions: SignOp<PUB>.() -> Unit = {},
```
babisRoutis, line 84: Assertions is not a concern of the JwtBatchSigner

babisRoutis (changes requested):
Please check my comments. More important

- KeyAttestation cannot be used for both JwtProofs and Attestation
- Assertions are not a concern of the Jwt Signers. Can be moved to the calling side

src/main/kotlin/eu/europa/ec/eudi/openid4vci/Signers.kt
vafeini: Removed assertions from construction of JwtSigner and JwtBatchSinger

src/main/kotlin/eu/europa/ec/eudi/openid4vci/Types.kt
```diff
@@ -240,6 +240,8 @@ sealed interface JwtBindingKey {
     }
 }
 
+class KeyAttestation
```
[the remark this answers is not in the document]
vafeini: Will revisit in the PR for Jwt Proofs with Key Attestations

babisRoutis (approved):

PR #426, opened 2025-06-23
Removed Oidc server metadata support

### LINKED ISSUE #424: Lookup Autherization Server Metadata using only `/.well-known/oauth-authorization-server`
Currently library checks for authorization server metadata (of the credential issuer) to 

- `/.well-known/oauth-authorization-server`
- `/.well-known/openid-configuration`

We should keep only the OAUTH2 option without lookup deviations

atsant:
Closes #424

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,7 +70,7 @@ As a wallet/caller having an out-of-band knowledge of a credential issuer,
 use the library to initiate issuance:
 
 - Fetch and validate the Credential Issuer Metadata
-- Fetch and validate the OAUTH2 or OIDC metadata used by the Credential Issuer
+- Fetch and validate the OAuth 2.0 Authorization Server metadata used by the Credential Issuer
 - Ensure that out-of-band knowledge is valid
 
 This is equivalent to [resolving a credential offer](#resolve-a-credential-offer), having authorization code grant, without issuer state.
@@ -128,7 +128,7 @@ This resolution includes the following
 - Check and validate the structure of the URI
 - Fetch the actual contents of the offer, in case URI points by reference to the offer
 - Fetch and validate the Credential Issuer Metadata
-- Fetch and validate the OAUTH2 or OIDC metadata used by the Credential Issuer
+- Fetch and validate the OAuth 2.0 Authorization Server metadata used by the Credential Issuer
 - Ensure that credential offer is aligned to both the above metadata & wallet issuance configuration
 
 #### Resolve a credential offer successful outcome:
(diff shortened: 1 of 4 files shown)


src/main/kotlin/eu/europa/ec/eudi/openid4vci/internal/DefaultAuthorizationServerMetadataResolver.kt
```diff
@@ -91,41 +63,24 @@ internal class DefaultAuthorizationServerMetadataResolver(
...
-
 /**
  * Gets a well-known metadata URL of an issuer.
  *
  * @receiver the Issuer
  * @param metadata the well-known metadata URL to get
- * @param lookup whether to produce a spec-compliant URL or not
  */
-private fun HttpsUrl.metadataUrl(metadata: String, lookup: Lookup): Url {
+private fun HttpsUrl.metadataUrl(metadata: String): Url {
```
babisRoutis, line 72: Please rename parameter `metadata` to `wellKnownPath`
Also rename function `metadadataUrl` to `formWellKnownUrl`or `wellKnownUrl`


Introduce a local (to the above function) function 

fun String.removePrefixAndSuffix(s: CharSequence): String = removePrefix(s).removeSuffix(s)


Assemble the pathSegment fully using the builder (now some parts are created using the builder and then there is a concatenation)


```kotlin
val pathSegment = buildString{
    append("/${wellKnownPath.removePrefixAndSuffix("/")}")
    if (joinedSegments.isNotBlank()) {
        append("/")
    }
    append(joinedSegments)
}


return URLBuilder(url).apply { path(pathSegment) }.build()

```

atsant:
@babisRoutis /@dzarras  Can you have a look if the impl looks ok and tests are good to go? 
Thanks in advance.

src/test/kotlin/eu/europa/ec/eudi/openid4vci/DefaultAuthorizationServerMetadataResolverTest.kt
```diff
@@ -131,7 +77,7 @@ internal class DefaultAuthorizationServerMetadataResolverTest {
 
         // Verify the last URL that was tried, is the common lookup for oauth2 authorization server metadata.
         assertEquals(
```
dzarras, line 79: @atsant Please rename this test case to something more meaningful. We no longer check a set of possible urls. We check a single url.

src/test/kotlin/eu/europa/ec/eudi/openid4vci/RequestMockers.kt
```diff
@@ -86,7 +81,7 @@ enum class IssuerMetadataVersion {
 }
 
 internal fun authServerWellKnownMocker(): RequestMocker = RequestMocker(
-    requestMatcher = endsWith("/.well-known/openid-configuration", HttpMethod.Get),
+    requestMatcher = endsWith("/.well-known/oauth-authorization-server", HttpMethod.Get),
```
dzarras, line 84: @atsant The `endWith` matcher is not 100% correct. This will be correct only when the URL of the Authorization Server does not contain a Path component. You should probably replace this with a custom matcher that checks whether the URL contains the well known Path segment.

src/test/kotlin/eu/europa/ec/eudi/openid4vci/RequestMockers.kt
```diff
@@ -86,7 +81,7 @@ enum class IssuerMetadataVersion {
 }
 
 internal fun authServerWellKnownMocker(): RequestMocker = RequestMocker(
```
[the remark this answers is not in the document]
dzarras: @atsant Disregard the previous comment.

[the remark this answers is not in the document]
dzarras: @atsant The method I mentioned returns Credential Issuer Metadata. Could you please rename the method `oiciWellKnownMocker()` to `credentialIssuerMetadataWellKnownMocker()` to avoid future confusion?

dzarras (approved):

dzarras (approved):

PR #431, opened 2025-07-11
Configure a Signer<JWK> as dpop signer instead of PopSigner.Jwt

### LINKED ISSUE #430: Deprecate PopSigner interface
After completion of #421 the creation of Proofs of type JWT will be done using `BatchSigner` and `Signer` provided by the wallet. This renders PopSigner interface redundant and can be deprecated. Currently `PopSigner`s are used during issuance requests for:

1. Generating PoP during issuance requests 
2. Generating DPoP signed jwts when the interacting authorization server supports that.

Τhe first case will be closed after #419 is merged. 

**NOTE:** There is still a dependency to PopSigner in the public api until the deprecated `RequestIssuance.request(IssuanceRequestPayload, List<PopSigner>)` is removed.

For the second case the following need to be done:
- [x] Change `OpenId4VCIConfig.dPoPSigner` property to be a `Signer<JWK>` 
- [x] Change `DeferredIssuerConfig.dPoPSigner` property to be a `Signer<JWK>` 
- [x] Change `DPoPJwtFactory` to work with 'Signer<JWK>' instead of `PopSigner`

vafeini:
Closes #430

### DIFF
--- a/src/test/kotlin/eu/europa/ec/eudi/openid4vci/examples/DeferredIssuerExtensions.kt
+++ b/src/test/kotlin/eu/europa/ec/eudi/openid4vci/examples/DeferredIssuerExtensions.kt
@@ -35,7 +35,7 @@ import kotlin.time.toKotlinDuration
 suspend fun DeferredIssuer.Companion.queryForDeferredCredential(
     clock: Clock = Clock.systemDefaultZone(),
     ctxTO: DeferredIssuanceStoredContextTO,
-    recreatePopSigner: ((String) -> PopSigner.Jwt)? = null,
+    recreatePopSigner: ((String) -> Signer<JWK>)? = null,
     recreateClientAttestationPodSigner: ((String) -> JWSSigner)? = null,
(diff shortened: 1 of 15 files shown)


src/test/kotlin/eu/europa/ec/eudi/openid4vci/examples/DeferredIssuerExtensions.kt
```diff
@@ -35,7 +35,7 @@ import kotlin.time.toKotlinDuration
 suspend fun DeferredIssuer.Companion.queryForDeferredCredential(
     clock: Clock = Clock.systemDefaultZone(),
     ctxTO: DeferredIssuanceStoredContextTO,
-    recreatePopSigner: ((String) -> PopSigner.Jwt)? = null,
+    recreatePopSigner: ((String) -> Signer<JWK>)? = null,
```
babisRoutis, line 38: Minor

There could be two cases to get the signer back

From a kid (The DPoP signer key is managed)
From a private JWK  (The DPoP signer key is ephemeral)

babisRoutis (approved):

PR #434, opened 2025-07-15
Add support for Key Attestations in JWT proofs

### LINKED ISSUE #377: Support key attestations in JWT proofs
TBD

### LINKED ISSUE #379: Support `attestation` proof
Draft 15 intoruces the concept of "key attestation" as a way for the holder to provide evidence for the authenticity and security properties of a key and its storage component to the Credential Issuer. 

Spec forsees that a key attestation jwt can be used as proof of possession of key material to be embedded in issued credentials (see [here](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0-15.html#section-8.2.1.3)).
When used as proof of possesion, a key attestation jwt MUST also include a fresh nonce value retrieved from issuer's nonce endpoint (see #372)

For the above the following are needed:

1. - [ ] Extend `Proof` sealed inerface to include the new `Attestation` type. In will contain internaly a `KeyAttestation` (the `KeyAttestation` construct will be introduced from #377)
2. Change the way `PopSigner` interface is used from callers of the issuance interface

Currently the proofs that need to be created by the library for an issuance request are dictated from the type of `PopSigner` that the caller provides. This is because so far the only thing needed from caller side was the key to sign the proof. Based on the type of the signer the appropriate proof is constructed internally from the library.

With key attestations used as proofs the key is not the only thing needed to create the proof, library needs more input from the caller to construct the proof (array of attested_keys, user_authentication, key_storage etc.). 

- [ ] PopSigner should be replaced with a new construct (ex. `ProofSpecification`) that depending on the case defines the following: 
	- [MANDATORY] A PopSigner as the carrier of all info needed for signing. 
	- [OPTIONAL] A proof payload (claimset) builder or function to give the caller a hook to pass the claims to be included in the generated proof 

- [ ] Proof Factory should use the ProofSpecification to build the proof accordingly: 
	- It will use the ProofSpecification's claimset builder to include the fresh nonce rerieved from nonce endpoint and then build the jwt's claimset 
	- It will use the ProofSpecification's signer to sign the key attestation. 
	- Specifically for the case of key attestation as proof the passed signer should be validated that has any type of bindingKey exept `JwtBindingKey.Attestation`

- [ ] PopSigner must stop to be used for deducting the type of proof to be created and be only used for singing proofs. For example see [here](https://github.com/eu-digital-identity-wallet/eudi-lib-jvm-openid4vci-kt/blob/main/src/main/kotlin/eu/europa/ec/eudi/openid4vci/internal/ProofBuilder.kt#L120)

gdimtsas:
Closes #377 
Closes #379


src/main/kotlin/eu/europa/ec/eudi/openid4vci/internal/RequestIssuanceImpl.kt
```diff
@@ -83,15 +67,34 @@ internal class RequestIssuanceImpl(
...
-                is ProofsSpecification.NoProofs -> buildRequest(requestPayload, null, credentialIdentifiers.orEmpty())
+                is ProofsSpecification.NoProofs -> {
+                    require(credentialConfigId.proofsRequirement() is CredentialProofsRequirement.ProofNotRequired) {
+                        "Proofs are required for credential configuration $credentialConfigId, but proofs specification is set to NoProofs"
+                    }
+                    buildRequest(requestPayload, null, credentialIdentifiers.orEmpty())
+                }
                 is ProofsSpecification.JwtProofs.NoKeyAttestation -> {
+                    ensure(
+                        credentialConfigId.keyAttestationProofsRequirement() !is KeyAttestationRequirement.Required &&
```
vafeini, line 78: Call credentialConfigId.keyAttestationProofsRequirement() once outside ensure

src/main/kotlin/eu/europa/ec/eudi/openid4vci/internal/RequestIssuanceImpl.kt
```diff
@@ -83,15 +67,34 @@ internal class RequestIssuanceImpl(
...
             when (proofsSpec) {
-                is ProofsSpecification.NoProofs -> buildRequest(requestPayload, null, credentialIdentifiers.orEmpty())
+                is ProofsSpecification.NoProofs -> {
+                    require(credentialConfigId.proofsRequirement() is CredentialProofsRequirement.ProofNotRequired) {
+                        "Proofs are required for credential configuration $credentialConfigId, but proofs specification is set to NoProofs"
+                    }
+                    buildRequest(requestPayload, null, credentialIdentifiers.orEmpty())
+                }
                 is ProofsSpecification.JwtProofs.NoKeyAttestation -> {
+                    ensure(
```
vafeini, line 77: This can be re-phrased to  
ensure(credentialConfigId.keyAttestationProofsRequirement() is KeyAttestationRequirement.NotRequired) { ... }

src/main/kotlin/eu/europa/ec/eudi/openid4vci/internal/RequestIssuanceImpl.kt
```diff
@@ -83,15 +67,34 @@ internal class RequestIssuanceImpl(
...
+                    }
+
                     proofsSpec.proofsSigner.use { signOps ->
                         val javaAlgorithm = proofsSpec.proofsSigner.javaAlgorithm
                         val proofsFactory = jwtProofsFactoryFrom(signOps, javaAlgorithm, credentialConfigId)
                         buildRequest(requestPayload, proofsFactory, credentialIdentifiers.orEmpty())
                     }
                 }
-                is ProofsSpecification.JwtProofs.WithKeyAttestation -> error("JWT Proofs with key attestation not yet supported")
+                is ProofsSpecification.JwtProofs.WithKeyAttestation -> {
```
vafeini, line 90: Is missing the respective check on the key attestation requirement. In this case key attestation requirement should be anything else than KeyAttestationRequirement.NotRequired

src/main/kotlin/eu/europa/ec/eudi/openid4vci/internal/RequestIssuanceImpl.kt
```diff
@@ -111,6 +114,15 @@ internal class RequestIssuanceImpl(
         }
     }
 
+    private fun CredentialConfigurationIdentifier.keyAttestationProofsRequirement(): KeyAttestationRequirement {
+        val credentialConfiguration = credentialSupportedById(this)
+        val spec = credentialConfiguration.proofTypesSupported.values.filterIsInstance<ProofTypeMeta.Jwt>().firstOrNull()
+        return when {
+            spec == null -> KeyAttestationRequirement.NotRequired
```
vafeini, line 121: I think this is misleading for the caller. In reality, there is no requirement because the JWT proof is not supported.  
If the JWT proof is not supported, then we cannot create proofs as we only support JWT proofs. 

This will change when the attestation proofs comes into play and we need to re-thing the whole proof generation/evaluation strategy

src/main/kotlin/eu/europa/ec/eudi/openid4vci/internal/RequestIssuanceImpl.kt
```diff
@@ -83,15 +67,34 @@ internal class RequestIssuanceImpl(
         // Place the request
         val (outcome, newResourceServerDpopNonce) = placeIssuanceRequest(accessToken, resourceServerDpopNonce) {
             when (proofsSpec) {
```
vafeini, line 69: I would start from what the proof requirement is based on the configuration (the CredentialProofsRequirement) and then see if it matches with the passed ProofsSpecification. 
Down the lines we also need to check the KeyAttestationRequirement of a specific proof type that is also deducted from the configuration. 
An idea is to make credentialConfigId.proofsRequirement() return all the info that we need about what are the requirements of the issuer about the specific proof type of this credential. We can enhance  CredentialProofsRequirement hierarchy with all the info we need (regarding requirements) and then as a second step compare it with what the caller passed as ProofsSpecification.

src/main/kotlin/eu/europa/ec/eudi/openid4vci/internal/RequestIssuanceImpl.kt
```diff
@@ -66,36 +71,53 @@ internal class RequestIssuanceImpl(
...
-                is ProofsSpecification.JwtProofs.NoKeyAttestation -> {
-                    ensure(
-                        credentialConfigId.keyAttestationProofsRequirement() !is KeyAttestationRequirement.Required &&
-                            credentialConfigId.keyAttestationProofsRequirement() !is KeyAttestationRequirement.RequiredNoConstraints,
-                    ) {
-                        CredentialIssuanceError.ProofTypeKeyAttestationRequired()
-                    }
+            val proofsRequirement = credentialConfigId.proofsRequirement()
+
+            var proofsFactory: ProofsFactory?
```
vafeini, line 76: Not needed since proofsFactory is created and used before calling buildRequest.

src/main/kotlin/eu/europa/ec/eudi/openid4vci/internal/RequestIssuanceImpl.kt
```diff
@@ -215,9 +243,9 @@ internal class RequestIssuanceImpl(
     }
 
     private suspend fun CredentialProofsRequirement.ProofRequired.cNonce(): CNonce? =
-        when (this) {
-            CredentialProofsRequirement.ProofRequired.WithoutCNonce -> null
-            CredentialProofsRequirement.ProofRequired.WithCNonce -> {
+        when (this.nonceRequirement) {
```
vafeini, line 246: 'this.' can be skipped

src/main/kotlin/eu/europa/ec/eudi/openid4vci/internal/JwtProofSigners.kt
```diff
@@ -69,3 +70,9 @@ internal fun JsonObjectBuilder.jwtProofHeader(key: JwtBindingKey) {
         }
     }
 }
+
+internal fun JsonObjectBuilder.keyAttestationJwtProofHeader(keyAttestation: KeyAttestationJWT, keyIndex: Int) {
+    put("typ", "openid4vci-proof+jwt")
```
babisRoutis, line 75: Minor comment

Create an object named OpenId4VCI
Declare const values for `openid4vci-proof+jwt` and `key_attestation`

src/main/kotlin/eu/europa/ec/eudi/openid4vci/KeyAttestationJWT.kt
```diff
@@ -0,0 +1,62 @@
...
+import com.nimbusds.jwt.SignedJWT
+
+data class KeyAttestationJWT(val value: String) {
+
+    val attestedKeys: List<JWK>
+
+    init {
+        val jwt = SignedJWT.parse(value)
+        jwt.ensureSignedNotMAC()
+        require(jwt.header.type != null && jwt.header.type.type.equals(KEY_ATTESTATION_JWT_TYPE)) {
```
babisRoutis, line 28: Use `==`

for instance

```kotlin
require(KEY_ATTESTATION_JWT_TYPE == jwt.header?.type?.type) {}

```

src/main/kotlin/eu/europa/ec/eudi/openid4vci/KeyAttestationJWT.kt
```diff
@@ -0,0 +1,62 @@
...
+            "Invalid Key Attestation JWT. Type must be set to `$KEY_ATTESTATION_JWT_TYPE`"
+        }
+        requireNotNull(jwt.jwtClaimsSet.issueTime) { "Invalid Key Attestation JWT. Misses `iat` claim" }
+
+        val attestedKeysClaimEntries = jwt.jwtClaimsSet.getListClaim("attested_keys")
```
babisRoutis, line 33: Consider the following simplification

```kotlin
require(!attestedKeysClaimEntries.isNullOrEmpty())
```

src/main/kotlin/eu/europa/ec/eudi/openid4vci/KeyAttestationJWT.kt
```diff
@@ -0,0 +1,62 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package eu.europa.ec.eudi.openid4vci
+
+import com.nimbusds.jose.jwk.JWK
+import com.nimbusds.jwt.SignedJWT
+
+data class KeyAttestationJWT(val value: String) {
```
babisRoutis, line 21: This could be a `value class` and (kotlinx) serializable

src/main/kotlin/eu/europa/ec/eudi/openid4vci/KeyAttestationJWT.kt
```diff
@@ -0,0 +1,62 @@
...
+
+    val attestedKeys: List<JWK>
+
+    init {
+        val jwt = SignedJWT.parse(value)
```
babisRoutis, line 26: Put this in a function to the companion object

something like `private fun validateJwt(jwt: String): Result<List<JWK>>

src/main/kotlin/eu/europa/ec/eudi/openid4vci/KeyAttestationJWT.kt
```diff
@@ -0,0 +1,62 @@
...
+ * limitations under the License.
+ */
+package eu.europa.ec.eudi.openid4vci
+
+import com.nimbusds.jose.jwk.JWK
+import com.nimbusds.jwt.SignedJWT
+
+data class KeyAttestationJWT(val value: String) {
+
+    val attestedKeys: List<JWK>
```
babisRoutis, line 23: Make it a calculated attribute 


```kotlin
val attestedKeys : List<JWK> = validateJwt(value).getOrThrow()
```

src/main/kotlin/eu/europa/ec/eudi/openid4vci/internal/SigningOps.kt
```diff
@@ -80,9 +84,16 @@ internal fun String.toJoseECAlg(): JWSAlgorithm? = when (this) {
     else -> null
 }
 
-internal fun SignFunction.Companion.forJavaEcPrivateKey(
+internal fun String.toJoseRSAAlg(): JWSAlgorithm? = when (this) {
```
babisRoutis, line 87: Please remove this

babisRoutis: Yeap

src/main/kotlin/eu/europa/ec/eudi/openid4vci/KeyAttestationJWT.kt
```diff
@@ -0,0 +1,74 @@
...
+        fun validateJwt(value: String): Result<List<JWK>> = runCatching {
+            val jwt = SignedJWT.parse(value)
+            jwt.ensureSignedNotMAC()
+
+            require(jwt.header?.type?.type == OpenId4VPSpec.KEY_ATTESTATION_JWT_TYPE) {
+                "Invalid Key Attestation JWT. Type must be set to `$OpenId4VPSpec.KEY_ATTESTATION_JWT_TYPE`"
+            }
+
+            val claims = Json.parseToJsonElement(jwt.jwtClaimsSet.toString()).jsonObject
+            require(claims["iat"] != null) { "Invalid Key Attestation JWT. Misses `iat` claim" }
```
babisRoutis, line 48: Minor: 
```kotlin
requireNotNull(claims["iat")

// even better 
val iat = claims["iat"]
requireNonNull(iat) { }

```

babisRoutis (dismissed):
A very minor comment

src/main/kotlin/eu/europa/ec/eudi/openid4vci/internal/RequestIssuanceImpl.kt
```diff
@@ -53,61 +55,114 @@ internal class RequestIssuanceImpl(
...
         val updatedAuthorizedRequest = this.withResourceServerDpopNonce(newResourceServerDpopNonce)
         updatedAuthorizedRequest to outcome.toPub()
     }
 
-    override suspend fun AuthorizedRequest.request(
-        requestPayload: IssuanceRequestPayload,
-        proofsSpec: ProofsSpecification,
-    ): Result<AuthorizedRequestAnd<SubmissionOutcome>> = runCatching {
-        val credentialConfigId = requestPayload.credentialConfigurationIdentifier
+    private suspend fun AuthorizedRequest.buildProofs(
```
babisRoutis, line 73: Just use grant, instead of AuthorizedRequest

babisRoutis (approved):
Nice.

@gdimtsas Just add a deprecated authenticate implemented by calling acquire

PR #452, opened 2025-08-16
Encrypted Requests

### LINKED ISSUE #442: Encrypted issuance and deferred issuance requests
The following apply to issuance and deferred issuance requests:

- [x] Issuance requests must be encrypted by the spec provided [here](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0-16.html#section-11.2.4-2.7.1) 
- [x] Encryption must be done according to section [9.4](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0-16.html#section-9.4) of the spec.
- [x] Investigate how compression can be applied to ecrypted requests before encrypting respecting `zip_values_supported` specified from issuer.

For issuance and deferred issuance requests, spec mentions the following

> Credential Request encryption MUST but used if the credential_response_encryption parameter is included, to prevent it being substituted by an attacker.

For this to happen when issuer includes in its metadata `credential_response_encryption` it must also provide `credential_request_encryption`. In other words if issuer supports encrypted responses it must advertice in metadata its capabilities for encrypted requests.

On the other hand spec does not seem to enforce response encryption when request encryption is required. Its seems that `credential_request_encryption` can be adverticed by issuer without providing `credential_response_encryption` metadata. 

- [x] Library should consider issuer metadata valid if `credential_response_encryption` is not included but `credential_request_encryption` is.
- [x] Library should consider issuer metadata as invalid if `credential_response_encryption` is included but no `credential_request_encryption`.
- [x] Requests (deferred or not) must be encrypted if the `credential_response_encryption` parameter is included in the request.
- [x] Requests (deferred or not) must be encrypted when issuer metadata have `credential_request_encryption.encryption_required` set to true

Based on section section [9.4](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0-16.html#section-9.4) of the spec, regarding encryption algorithm the following is specified

> The JWE alg algorithm used MUST be equal to the alg value of the chosen JWK.

This mandates that the JWKs used for encryption (either wallet generated or issuer adverticed) must include an 'alg' attribute.

- [x] JWKs included in issuer's `credential_request_encryption' must be validated to have `alg` specified
- [x] JWKs generated from lib to be used for encryption must include the encryption algorithm in `alg` jwk attribute

vafeini:
Closes #442


src/main/kotlin/eu/europa/ec/eudi/openid4vci/internal/http/CredentialIssuerMetadataJsonParser.kt
```diff
@@ -701,6 +701,13 @@ private fun CredentialIssuerMetadataTO.toDomain(expectedIssuer: CredentialIssuer
     val credentialRequestEncryption = credentialRequestEncryption.toDomain()
     val credentialResponseEncryption = credentialResponseEncryption.toDomain()
 
+    // If issuer supports or requires credential response encryption then it must advertise its request encryption capabilities
+    if (credentialResponseEncryption !is CredentialResponseEncryption.NotSupported) {
```
babisRoutis, line 705: IMO, we are repeating the same check (the association between request and response encryption) in two different places.

The one is right here and the other to the class `IssuanceEncryptionSpecs`

src/main/kotlin/eu/europa/ec/eudi/openid4vci/Types.kt
```diff
@@ -270,6 +291,39 @@ data class IssuanceResponseEncryptionSpec(
...
+ * Each encryption specification leverages the `EncryptionSpec` class to ensure compatibility and secure encryption using standards
+ * such as JWKs and JWE algorithms.
+ *
+ * @property requestEncryptionSpec The encryption specification for securing issuance requests sent to the issuer.
+ * @property responseEncryptionSpec The encryption specification for securing issuance responses returned by the issuer.
+ *
+ * @throws MissingRequiredRequestEncryptionSpecification Thrown when a response encryption specification is provided
+ * but request encryption specification is missing.
+ */
+data class IssuanceEncryptionSpecs(
```
babisRoutis, line 314: From the description of the class, it is not clear to me what exactly it represents.

- The choices (or requirements) of the Credential Issuer?
- Some kind of wallet policy?
- The "agreement between wallet-credential issuer"?

src/main/kotlin/eu/europa/ec/eudi/openid4vci/internal/IssuanceEncryption.kt
```diff
@@ -0,0 +1,210 @@
...
+ * @param credentialRequestEncryption Issuer metadata about its capability to encrypt issuance requests.
+ * @param credentialResponseEncryption Issuer metadata about its capability to encrypt issuance responses.
+ * @param responseEncryptionSpecFactory A factory for creating the response encryption specification based on supported parameters.
+ * @param requestEncryptionSpecFactory A factory for creating the request encryption specification needed for secure communication.
+ * @return A Result object containing the `IssuanceEncryptionSpecs` that hold the encryption specifications
+ *         for request and response, or an error if encryption specification creation fails.
+ */
+internal fun issuanceEncryptionSpecs(
+    encryptionSupportConfig: EncryptionSupportConfig,
+    credentialRequestEncryption: CredentialRequestEncryption,
```
babisRoutis, line 36: Consider replacing the parameters `credentialRequestEncryption` & `credentialResponseEncryption` with a single parameter `credentialIssuerMetadata`.

The `CredentialIssuerMetadata` offers some guarantees that we don't use two arbitrary request and response encryption params, given that there is an association  between them.

src/main/kotlin/eu/europa/ec/eudi/openid4vci/Issuance.kt
```diff
@@ -176,7 +175,10 @@ interface RequestIssuance {
  * that the wallet expects in the response of its issuance request.
  */
 typealias ResponseEncryptionSpecFactory =
-    (SupportedResponseEncryptionParameters, KeyGenerationConfig, List<CompressionAlgorithm>?) -> IssuanceResponseEncryptionSpec?
+    (SupportedResponseEncryptionParameters, EncryptionSupportConfig) -> EncryptionSpec?
```
babisRoutis, line 178: Perhaps, we should replace those two typealias(es) with two functional interfaces.

To the companion objects of these interface we can move the factory methods now located to `Issuer.kt`

src/main/kotlin/eu/europa/ec/eudi/openid4vci/Issuer.kt
```diff
@@ -307,24 +309,50 @@ interface Issuer :
...
+                    .intersect(walletEncryptionSupportConfig.supportedEncryptionMethods).firstOrNull()
+                encryptionMethod?.let { method ->
+                    issuerSupportedResponseEncryptedParameters.algorithms.firstNotNullOfOrNull { algorithm ->
+                        KeyGenerator.genKeyIfSupported(walletEncryptionSupportConfig, algorithm)
+                            ?.let { jwk -> EncryptionSpec(jwk, method, compressionAlg) }
                     }
                 }
             }
+
+        val DefaultRequestEncryptionSpecFactory: RequestEncryptionSpecFactory =
```
babisRoutis, line 336: Currently, `RequestEncryptionSpecFactory`is a typealias. Should we make it a fun interface, we can move to the companion object the code bellow as a factory method

src/main/kotlin/eu/europa/ec/eudi/openid4vci/Types.kt
```diff
@@ -239,13 +241,32 @@ sealed interface JwtBindingKey {
...
+ * Represents the specifications and parameters required for encryption. Used for encrypting issuance requests or for communicating
+ * to issuers the encryption parameters by which expected issuance responses should be encrypted by.
+ *
+ * This class is designed to encapsulate encryption-related configurations,
+ * including the key, encryption method, and an optional compression algorithm.
+ * It validates the provided key and algorithm for compatibility with the
+ * encryption operation and ensures that they adhere to the expected asymmetric
+ * encryption standards.
+ *
+ * @property jwk The JSON Web Key (JWK) used for encryption, representing the cryptographic key.
```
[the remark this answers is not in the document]
vafeini: The `EncryptionSpec` data class is used to describe both the request and the response encryption parameters. In the case of request, jwk is the issuer's encryption key that the wallet selected. In the case of response, jwk is the wallet's key that the issuer will use to encrypt the response.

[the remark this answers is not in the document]
babisRoutis: Dear @vafeini 

Your description is correct, but my point was  about the name of the attribute.
I believe that `recipientKey` is more descriptive of the usage of the key, rather than the neutral term `jwk`.

Indeed, to the request the `recipientKey` is the issuer's key (that's the party receiving the encrypted request) and in the response the recipient is the wallet.

dzarras (dismissed):

src/main/kotlin/eu/europa/ec/eudi/openid4vci/DeferredIssuer.kt
```diff
@@ -30,22 +34,17 @@ import java.time.Clock
...
  * @param dPoPSigner the signer that was used for DPoP. Must be provided only if DPoP was used.
- * @param responseEncryptionSpec the encryption key, used for credential response encryption. Used only
- * it was included in the initial credential request
  * @param clock Wallet's clock
  */
 data class DeferredIssuerConfig(
     val credentialIssuerId: CredentialIssuerId,
     val client: Client,
-    val deferredEndpoint: URL,
     val authServerId: URL,
```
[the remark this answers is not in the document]
vafeini: Removed re-negotiation of encryption specs. The required data are now stored in `DeferredIssuerConfig`.  This includes 
- the request encryption spec as it was negotiated in the initial request
- the response encryption's compression algorithm and encryption method as negotiated in the initial request

PR #465, opened 2025-09-15
Update Attestation-Based Client Authentication support (Draft 7)

### LINKED ISSUE #376: Update Wallet Attestation support
Currently library supports wallet attestations by implementing 'OAUTH2 Attestation-Based Client Authentication' draft 03

- [ ] Draft 04 is published that should be evaluated for updates possibly needed.
- [ ] OpenID4VCI draft 15 specifies the structure that a wallet attestations should have ([Appendix E](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0-15.html#appendix-E)). Library defines data class `ClientAttestationJWT` as a holder of `SignedJwt` that represents a wallet attestation jwt and perfoms some sanity checks about its structure. These checks should be evaluated and updated according to the structure defined by the spec.

dzarras:
This PR updates Attestation-Based Client Authentication support to **[Draft 07](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-attestation-based-client-auth-07)**

Closes #376

### DIFF
--- a/src/main/kotlin/eu/europa/ec/eudi/openid4vci/ClientAttestation.kt
+++ b/src/main/kotlin/eu/europa/ec/eudi/openid4vci/ClientAttestation.kt
@@ -20,6 +20,7 @@ import com.nimbusds.jose.JWSAlgorithm
 import com.nimbusds.jose.JWSObject
 import com.nimbusds.jose.crypto.MACSigner
 import com.nimbusds.jose.jwk.JWK
+import com.nimbusds.jose.util.JSONObjectUtils
 import com.nimbusds.jwt.SignedJWT
 import eu.europa.ec.eudi.openid4vci.internal.*
 import kotlinx.serialization.Required
@@ -82,6 +83,11 @@ value class ClientAttestationPoPJWT(val jwt: SignedJWT) {
     }
 
     val clientId: ClientId get() = checkNotNull(jwt.jwtClaimsSet.issuer)
+    val claims: ClientAttestationPOPClaims
+        get() =
+            JsonSupport.decodeFromString<ClientAttestationPOPClaims>(
+                JSONObjectUtils.toJSONString(jwt.jwtClaimsSet.toJSONObject()),
+            )
 }
 
 @Serializable


src/main/kotlin/eu/europa/ec/eudi/openid4vci/ClientAttestation.kt
```diff
@@ -45,19 +44,21 @@ typealias ClientAttestation = Pair<ClientAttestationJWT, ClientAttestationPoPJWT
...
-        val cnf = requireNotNull(jwt.jwtClaimsSet.cnf()) { "Invalid Attestation JWT. Misses cnf claim" }
-        requireNotNull(cnf.cnfJwk()) { "Invalid Attestation JWT. Misses jwk claim from cnf" }
-        requireNotNull(jwt.jwtClaimsSet.expirationTime) { "Invalid Attestation JWT. Misses exp claim" }
+        jwt.ensureType(JOSEObjectType(AttestationBasedClientAuthenticationSpec.ATTESTATION_JWT_TYPE))
+        requireNotNull(jwt.jwtClaimsSet.issuer) { "Invalid Attestation JWT. Misses `iss` claim" }
+        requireNotNull(jwt.jwtClaimsSet.subject) { "Invalid Attestation JWT. Misses `sub` claim" }
+        requireNotNull(jwt.jwtClaimsSet.expirationTime) { "Invalid Attestation JWT. Misses `exp` claim" }
+        val cnf = requireNotNull(jwt.jwtClaimsSet.cnf()) { "Invalid Attestation JWT. Misses `cnf` claim" }
+        requireNotNull(cnf.cnfJwk()) { "Invalid Attestation JWT. Misses `jwk` claim from `cnf`" }
+        jwt.ensureSignedOrVerified()
```
gdimtsas, line 56: According to [9. Verification and Processing](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-attestation-based-client-auth-07#name-verification-and-processing), we must also require that the used algorithm is asymmetric.

> The alg JOSE Header Parameter for both JWTs indicates a registered asymmetric digital signature algorithm

dzarras: Yet in https://datatracker.ietf.org/doc/html/draft-ietf-oauth-attestation-based-client-auth-07#name-client-attestation-jwt they mention:

> The JWT MUST be digitally signed or integrity protected with a Message Authentication Code (MAC). The authorization server MUST reject JWTs if signature or integrity protection validation fails.

Additionally in https://datatracker.ietf.org/doc/html/draft-ietf-oauth-attestation-based-client-auth-07#name-document-history:

> remove restrictions to not allow MAC-based algorithms

I suspect the authors haven't updated the section you mention.

gdimtsas: You're right, it seems an oversight on the spec.

babisRoutis: @dzarras , @gdimtsas Perhaps you could open an issue to the spec. If so, make sure you check the open issues firsts

src/main/kotlin/eu/europa/ec/eudi/openid4vci/ClientAttestation.kt
```diff
@@ -84,17 +89,10 @@ value class ClientAttestationPoPJWT(val jwt: SignedJWT) {
 
 data class ClientAttestationPoPJWTSpec(
     val signingAlgorithm: JWSAlgorithm,
```
babisRoutis, line 91: Why `duration` was removed?

babisRoutis:
> > @dzarras
> > Irrelevant to the ABCA changes:
> > 
> > * perhaps it is a good opportunity to rename sealed hierarchy `Client` to `ClientAuthentication`
> > * A client that uses `none` authentication method may have a `ClientId`
> > * Do we really want `CNonce` value class? Perhaps a `Nonce` value class makes more sense (Nimbus provides something similar but there is no need to depend on it)
> 
> @babisRoutis
> 
>     1. Do the following names make sense?
> 
> 
> ```kotlin
> sealed interface ClientAthentication {
>    data class None(...) : ClientAuthnetication
>    data class AttestationBased(...) : ClientAuthnetication
> ```


None is more correct than Public. So we could use this

PR #468, opened 2025-09-26
Replace runCatching

### LINKED ISSUE #437: Replace runCatching
Replace `runCatching{ }` with 

```kotlin
internal inline fun <R> runCatchingCancellable(block: () -> R): Result<R> {
    return try {
        Result.success(block())
    } catch (ce: CancellationException) {
        throw ce
    } catch (e: Exception) {
        Result.failure(e)
    }
}
```

atsant:
This PR replaces usages of `runCatching` in suspend functions with `runCatchingCancellable`. `runCatchingCancellable` is a variant of `runCatching` that always propagates `CancellationException` to the caller.

Closes #437


atsant:
`runCatchingCancellable` is now catching `Τhrowable` instead of `Exception` unlike implementations in other libraries.

Maybe we need to change Throwables inside the code base to Exceptions.

dzarras (approved):

PR #502, opened 2026-04-01
Implemented credential reuse policies per credential configuration

### LINKED ISSUE #494: Support ETSI TS 119 472-3 - Attestation provider Reuse Policies
ETSI TS 119 472-3 defines specific requirements on how attestation providers should advertice their policies regarding how the issued credential should be re-used when presented to relying parties

The specification extends a credential issuer's metadata (as defined in [Openid4VCI section 12.2.4](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#section-12.2.4)) as follows:
- Attribute `credential_reuse_policy` may optionally be present in the `credential_metadata` child parameter of each credential configuration in `credential_configurations_supported`
- Attribute `credential_reuse_policy` must be a json object that has an `id` member that identifies the reuse policy. Its value must be a json string
- Attribute `credential_reuse_policy` may contain an `options` attrubute defining extra details about the specific reuse policy. It is an array of json objects. 

**_Important:_** Absence of the credential_reuse_policy parameter shall be understood as that the attestation provider does not limit the number of times that a Wallet Unit may present the PID/EAA.

**Reuse policies as defined in ARF** 
ARF in Annex II defines 4 methods about how a wallet unit can control the presentation of attestation. These methods directly affect the issuance of the attestion (batch or single instance) and the triggering event for the attestation's re-issuance.

**Method A - Once-only** 
	Dectates that attestations will be issued in batches. The size of the batch is controled by the specific policy's details. Wallet unit must present each instance of the attestation only once to a Relying Party. There is a lower limit for the number of unused attestations. When this limit is reached wallet unit must request the issuance of a new batch.

**Method B - Limited-time** 
	TBD

**Method C - Rotating-batch**  
	TBD

**Method D - Per-Relying Party**  
	TBD

| Method                | Issuance                                 | Presentation policy                                                                                                                                                 | Re-Issuance triggering event               |
|-----------------------|------------------------------------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------|--------------------------------------------|
| Once-only (A)         | Batch issuance, size controled by policy | Each instance presented only once                                                                                                                                   | Lower limit of unused attestations reached |
| Limited-time (B)      | Single instance for a limited period     | Same instance presented multiple times                                                                                                                              | Attestation expiration time is approaching |
| Rotating-batch (C)    | Batch issuance, size controled by policy | Present each instance in a batch once per Relying party in a random order. <br/> When all isntances presented once reset the batch and start over in randdom order. | Attestation expiration time is approaching |
| Per-Relying Party (D) | Batch issuance, size controled by policy | Present a different instance to each different Relying Party. <br/>Same instance should be presented to the same RP if RP request attesation multiple times.        | Attestation expiration time is approaching |

In the context of EUDIW the `credential_reuse_policy` is further specified to follow attestation reuse policies as they are defined is ARF Annex II:
- Value of the policy's `id` attribute must be `arf_annex_ii`
- Attribute `options` is mandatory to exist. 
- Each json object in `options` array must contain a `details` attribute that is an array of json strings.
- Allowed values of `details` are: `once_only`, `limited_time`, `rotating-batch`, `per-relying-party`
- Attribute `details` is mandatory to contain either `once_only` or `limited_time` (but not both of them) and additionaly it might contain `rotating-batch` or `per-relying-party` or both.
- If `details` array contains one of `once_only`, `rotating-batch`, or `per-relying-party` then the json object must also contain attribute `batch_size`. Its value is numeric and specifies the size of the batch during issuance.
- If `details` array contains `once_only` then the json object must also contain attribute `reissue_trigger_unused`. Its value is numeric, should be lower than `batch_size`. 
	It defines the lower limit for the number of unused attestations, that when reached wallet unit must request the issuance of a new batch.
- If `details` array contains one of `limited-time`, `rotating-batch`, or `per-relying-party` then the json object must also contain attribute `reissue_trigger_lifetime_left`. 
	Its value is bumeric and specifies the number of seconds before the expiration of the attestation. When this instant arrives wallet unit must request the re-issuance.
- When multiple policy options are defined (multiple elements in `options` array) they must not be overlapping, meaning that the valued in the respective details attribute must be unique.
- When multiple policy options are defined wallet must select the first that it can support

Indicative structure of credential_reuse_policy object.

```
{
	"credential_reuse_policy": {
		"id": "arf_annex_ii",
		"options" : [{
			"details" : ["once_only", "limited_time", "rotating-batch", "per-relying-party"],
			"batch_size": 5,						<-- applicable only to "once_only", "rotating-batch", or "per-relying-party"
			"reissue_trigger_unused": 2, 			<-- applicable only to "once_only" and should be lower than "batch_size",
			"reissue_trigger_lifetime_left": 655433	<-- applicable only to "limited-time", "rotating-batch", or "per-relying-party" and specifies the number of seconds before the expiration 
			
		}]
	}
}
```

Allowed combinations of values in the "details" array:
1. ["once_only"]
2. ["once_only", "rotating-batch"]
3. ["once_only", "per-relying-party"]
4. ["once_only", "rotating-batch", "per-relying-party"]
5. ["limited-time"] 
6. ["limited-time", "rotating-batch"] 
7. ["limited-time", "per-relying-party"] 
8. ["limited-time", "rotating-batch", "per-relying-party"]

gdimtsas:
Closes #494


src/main/kotlin/eu/europa/ec/eudi/openid4vci/internal/RequestIssuanceImpl.kt
```diff
@@ -242,7 +260,7 @@ internal class RequestIssuanceImpl(
             javaSigningAlgorithm.toSupportedJoseAlgorithm(credentialConfigId)
         }
         return proofsSpecification.proofsSigner.use { operation ->
-            operation.assertMatchesBatchIssuanceBatchSize()
+            operation.assertMatchesBatchIssuanceBatchSize(credentialConfigId)
```
vafeini: The check for compliance with reuse policy should be extended also to the other proof types (Attestation proof and jwt proof with key attestations)

src/main/kotlin/eu/europa/ec/eudi/openid4vci/CredentialReusePolicy.kt
```diff
@@ -0,0 +1,180 @@
...
+            "details must not contain both once_only and limited_time"
+        }
+
+        val requiresBatchSize = hasOnceOnly || hasRotatingBatch || hasPerRelyingParty
+        if (requiresBatchSize) {
+            requireNotNull(batchSize) { "batch_size is required when details contains once_only, rotating-batch, or per-relying-party" }
+            require(batchSize > 0) { "batch_size must be greater than 0" }
+        }
+
+        if (hasOnceOnly) {
```
vafeini: The `reissue_trigger_unused` should exist also in the case of  `per-relying-party` option. I have updated the analysis of the issue, it was missing. This is something that currently does not exist in the standard but will be added at some point.

src/main/kotlin/eu/europa/ec/eudi/openid4vci/CredentialReusePolicy.kt
```diff
@@ -0,0 +1,180 @@
...
+        private fun readResolve(): Any = None
+        override fun effectiveBatchSize(supportedReusePolicies: Set<SupportedReusePolicy>): Int? = null
+    }
+
+    /**
+     * ARF Annex II reuse policy.
+     *
+     * @param options extra details about the specific reuse policy
+     */
+    data class ArfAnnex2ReusePolicy(
```
vafeini: Re-use policies as defined in the standard can be defined either as multiple single element in `options` array with multiple values in `details` array or multiple elements in `options` array or combinations of the above. To alleviate this indirection and make the domain model more clear I would model the ArfAnnex2ReusePolicy as a sealed hierarchy like the one below. In CredentialReusePolicyTO.toDomain() I would do the appropriate convertion and end up with a list of valid `ArffAnnex2ReusePolicy`s.      


NOTE: Policies can be further categorized to `BatchPolicy`s and `SingleInstancePolicy`s  



This model is a cleaner for callers of the library as there might be cases that the library will need to associate an issued attestation with the policy used.  

```
sealed interface ArffAnnex2ReusePolicy : CredentialReusePolicy {

        data class LimitedTime(
            val reissueTriggerLifetimeLeft: Int,
        ): ArffAnnex2ReusePolicy

        data class OnceOnly(
            val batchSize: Int,
            val reissueTriggerUnused: Int,
        ): ArffAnnex2ReusePolicy {
            init {
                require(batchSize > 2) { "...." }
            }            
        }

        data class RotatingBatch(
            val batchSize: Int,
            val reissueTriggerUnused: Int,
        ): ArffAnnex2ReusePolicy {
             init {
                require(batchSize > 2) { "...." }
            }
        }

        data class PerRelyingParty(
            val batchSize: Int,
            val reissueTriggerUnused: Int,
            val reissueTriggerLifetimeLeft: Int,
        ): ArffAnnex2ReusePolicy {
             init {
                require(batchSize > 2) { "...." }
            }
        }
        
}
```

src/main/kotlin/eu/europa/ec/eudi/openid4vci/Config.kt
```diff
@@ -85,6 +86,7 @@ data class OpenId4VCIConfig(
     val parUsage: ParUsage = ParUsage.IfSupported,
     val clock: Clock = Clock.systemDefaultZone(),
     val issuerMetadataPolicy: IssuerMetadataPolicy = IssuerMetadataPolicy.IgnoreSigned,
+    val supportedCredentialReusePolicies: Set<SupportedReusePolicy> = emptySet(),
```
babisRoutis: Do this have to be a set?

Why not a data class with 4 nullable attributes?

dzarras:
@gdimtsas Please rebase against `release/0.11.0` when time permits.

vafeini (dismissed):

vafeini (dismissed):

src/main/kotlin/eu/europa/ec/eudi/openid4vci/internal/RequestIssuanceImpl.kt
```diff
@@ -253,30 +287,57 @@ internal class RequestIssuanceImpl(
...
+    private fun Int.assertMatchesBatchIssuanceBatchSize(
+        selectedReusePolicy: EudiReusePolicy?,
+    ) = when (this) {
+        0 -> error("At least one PopSigner is required in Authorized.ProofRequired")
+        1 -> Unit
+        else -> {
+            if (selectedReusePolicy != null) {
+                when (selectedReusePolicy) {
+                    is EudiReusePolicy.LimitedTime ->
+                        CredentialIssuanceError.IssuerDoesNotSupportBatchIssuance()
```
dzarras: @gdimtsas Shouldn't this be thrown?

src/main/kotlin/eu/europa/ec/eudi/openid4vci/internal/RequestIssuanceImpl.kt
```diff
@@ -253,30 +287,57 @@ internal class RequestIssuanceImpl(
...
+                    else -> {
+                        val policyBatchSize = selectedReusePolicy.batchSize
+                        ensureNotNull(policyBatchSize) {
```
dzarras: @gdimtsas I'd simplify this as:
```kotlin
 val policyBatchSize = checkNotNull(selectedReusePolicy.batchSize) {
    "Batch reuse policy ${selectedReusePolicy::class} with null batch size"
}
```

and avoid calling `error` inside `ensureNotNull`. 

`error` throws an `IllegalStateException`, while `ensureNotNull` expects a lambda that returns a `Throwable` to be thrown in case the provided value is `null`. 

Calling `error` inside `ensureNotNull` works, because it returns `Nothing`.

src/main/kotlin/eu/europa/ec/eudi/openid4vci/CredentialReusePolicy.kt
```diff
@@ -0,0 +1,241 @@
...
+}
+
+/**
+ * A single ARF Annex II option in the reuse policy.
+ */
+sealed interface EudiReusePolicy {
+
+    val batchSize: Int?
+    val reissueTriggerUnused: Int?
+    val reissueTriggerLifetimeLeft: Long?
```
dzarras: @gdimtsas I'd suggest to use `Duration` for `reissueTriggerLifetimeLeft`.

dzarras (changes requested):

dzarras (dismissed):

vafeini (dismissed):

vafeini (approved):

PR #503, opened 2026-04-01
Introduce an explict policy for DPoP usage

### LINKED ISSUE #504: Introduce an explict policy for DPoP usage
*Rationale*: `OpenId4VCIConfig` allows a Wallet to configure policies for PAR usage and Credential Issuer Signed Metadata usage, but not DPoP usage. 

*Issue*:
When it comes to DPoP, a Wallet simply provides a DPoP Proof signer, and if the Authorization Server supports DPoP, then the library enables and transparently handles DPoP. Otherwise it falls back to Bearer Token Authentication.
The later might not be a desirable behavior for a Wallet.

This approach doesn't allow a Wallet to require or never use DPoP if so desired.

*Recommended approach*:
Introduce `DPoPUsage`: a Wallet policy regarding DPoP usage.

This would enable a Wallet to:
1. Never use DPoP
2. Use DPoP if supported by the Authorization Server
3. Require DPoP usage; If the Authorization Server doesn't support DPoP, issuance cannot proceed

When a Wallet supports or requires DPoP, it provides a DPoP Proof signer, i.e. `Signer<JWK>`. 
* The JWS Algorithm used by the signer must be supported by the Authorization Server. 
* If the Authorization Server supports DPoP but does not support the JWS Algorithm used by the Wallet provided DPoP Proof signer, an error is raised.

dzarras:
Closes #504

### DIFF
--- a/README.md
+++ b/README.md
@@ -573,7 +573,7 @@ data class OpenId4VCIConfig(
     val authFlowRedirectionURI: URI,
     val encryptionSupportConfig: EncryptionSupportConfig,
     val authorizeIssuanceConfig: AuthorizeIssuanceConfig = AuthorizeIssuanceConfig.FAVOR_SCOPES,
-    val dPoPSigner: Signer<JWK>? = null,
+    val dPoPUsage: DPoPUsage = DPoPUsage.Never,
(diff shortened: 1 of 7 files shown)


babisRoutis (dismissed):
Just a couple of minor comments. Otherwise approved

PR #505, opened 2026-04-01
Improve guards around illegal states when using DPoP

### LINKED ISSUE #507: Improve handling of DPoP Access Tokens
When a DPoP Access Token is used, a DPoP Proof must be generated and sent alongside the DPoP Access Token.

CredentialEndpointClient, DeferredEndPointClient, and NotificationEndPointClient can be used either with a Bearer or a DPoP Access Token.

Currently, these clients generate and send a DPoP Proof when the Access Token is DPoP and a DPoPJwtFactory has been configured.
If either the Access Token is not DPoP, or a DPoPJwtFactory has not been configured, no DPoP Proof is generated.

This behavior is not correct. When a DPoP Access Token is used a DPoPJwtFactory must be provided. Not having a DPoPJwtFactory is an illegal state.

Finally, `HttpRequestBuilder.bearerOrDPoPAuth()` silently falls back to sending a DPoP Access Token as a Bearer Access Token when no DPoP Proof is provided. This is not correct and must not be allowed.

dzarras:
Closes #507 

This PR:
1. Updates the above clients to check that a `DPoPJwtFactory` is configured when using DPoP Access Tokens
2. Updates `HttpRequestBuilder.bearerOrDPoPAuth()` to check that a DPoP Proof is used when a DPoP Access Token has been provided, instead of silently falling back to sending the DPoP Access Token as a Bearer Token.
3. Marks `HttpRequestBuilder.bearerOrDPoPAuth()` as internal.
4. Updates `DeferredIssuanceContext` to ensure that a DPoP Signer is provided when the used Access Token is DPoP.

### DIFF
--- a/src/main/kotlin/eu/europa/ec/eudi/openid4vci/DeferredIssuer.kt
+++ b/src/main/kotlin/eu/europa/ec/eudi/openid4vci/DeferredIssuer.kt
@@ -78,8 +78,10 @@ data class DeferredIssuanceContext(
     val authorizedTransaction: AuthorizedTransaction,
 ) {
     init {
-        require(authorizedTransaction.authorizedRequest.accessToken !is AccessToken.DPoP || null != config.dPoPSigner) {
-            "config.dPoPSigner is required when DPoP access tokens are used"
+        if (authorizedTransaction.authorizedRequest.accessToken !is AccessToken.DPoP) {
+            requireNotNull(config.dPoPSigner) {
+                "config.dPoPSigner is required when DPoP access tokens are used"
+            }
         }
     }
 }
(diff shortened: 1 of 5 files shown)


src/main/kotlin/eu/europa/ec/eudi/openid4vci/DeferredIssuer.kt
```diff
@@ -76,7 +76,13 @@ data class AuthorizedTransaction(
 data class DeferredIssuanceContext(
     val config: DeferredIssuerConfig,
     val authorizedTransaction: AuthorizedTransaction,
-)
+) {
+    init {
+        require(authorizedTransaction.authorizedRequest.accessToken !is AccessToken.DPoP || null != config.dPoPSigner) {
```
babisRoutis: Very minor,

In general, I find more easily to to read that 
If accessToken is DPoP then requireNotNull dpopSigner

babisRoutis (changes requested):

babisRoutis (approved):

PR #512, opened 2026-04-08
Support only JWT Proof + Key Attestation, and Attestation Proof

### LINKED ISSUE #508: Introduce a configuration option to enforce the use of key_attestation
According to HAIP v1 and ETSI TS 119 472-3, proofs of type `jwt` cannot be used, unless they include a `key_attestation`  and signed with the first of the attested keys (found in the `key_attestation`).

This means that the wallet for device-bound attestations must use either:
- Proof of type `attestation` 
- Proof of type `jwt` + `key_attestation`

That is, Proof of type `jwt` (without `key_attestation`) is not allowed

In addition, HAIP v1 allows issuance of non-device-bound attestations, where Proofs are not used.

The above rules are not being taken into account by the present library when processing a credential offer. 
Library should reject a  credential configurations that require Proof of type `jwt` without `key_attestation`

For this reason we should
- Introduce a new option to allow caller define its policy with regards to non-device-bound and device-bound attestations. The former could be a boolean flag, where as the later could be a data class describing the algorithms supported for signing the KA and/or the Proof of type JWT+KA.

- Depending on the option, library should validate the credential configuration.

### LINKED ISSUE #509: Remove support for Proof of type `jwt` without `key_attestation`
Device-bound attestations require the presence of `key_attestation` to the credential request.

This means that the valid options for proofs are:
- Proof of type `attestation`
- Proof of type `jwt` with `key_attestation` on the protected header.

The library currently does not enforce this. In particular allows usage of Proofs with  type `jwt` without `key_attestation`

### LINKED ISSUE #520: Issuers metadata parsing updates to align with TS3 v1.5
TS3 in section 2.4 defined that issuer metadata should be extended with:

> The following field is to be placed within the key_attestations_required object as specified in https://github.com/eu-digital-identity-wallet/eudi-doc-standards-and-technical-specifications/issues/3:
> - preferred_key_storage_status_period: OPTIONAL. An integer specifying a PID or Attestation Provider's preference for the remaining status maintenance period (key_storage_status.exp minus current time) of the KA it receives during issuance, in seconds. See [Section 2.4.2](https://github.com/eu-digital-identity-wallet/eudi-doc-standards-and-technical-specifications/blob/main/docs/technical-specifications/ts3-wallet-unit-attestation.md#242-wallet-provider-responsibilities) for how the Wallet Unit and the Wallet Provider must handle this parameter.

- [ ] Accept attribute `preferred_key_storage_status_period` in `key_attestations_required` element of `jwt` and `attestation` proof types of `proof_types_supported` array.
   
TS3 in section 2.2.2.2 mentions:

> A PID Provider or an Attestation Provider issuing device-bound attestations SHALL indicate in the proof_types_supported parameter in its Issuer Credential Metadata (see https://github.com/eu-digital-identity-wallet/eudi-doc-standards-and-technical-specifications/issues/3) that it supports both the jwt and the attestation proof types for KAs, both of which SHALL include the key_attestation_required object.

- [ ] For key-bound credentials (have `proof_types_supported` in credential configuration) both `jwt` and `attestation` proof types must be supported and include `key_attestations_required` object. 

ΙΑ 2024/2979 mentions the following in Annex Ib:

> TR_KA-4.1: Where a credential issuer issues non-device-bound credentials, it shall omit the "proof_types_supported" and the "cryptographic_binding_methods_supported" parameters in the Credential Issuer Metadata.

- [ ] Either both `cryptographic_binding_methods_supported` and `proof_types_supported` should be present in a credential configuration or both missing

dzarras:
This PR updates the library to only support JWT Proofs that contain Key Attestation and Attestation Proofs.

Credential Issuers are now required to support both JWT Proofs with Key Attestation and Attestation Proofs per TS3/ETSI 119 472-3.

Batch Issuance is supported using either a single JWT Proof with a Key Attestation that attests multiple keys, or a single Attestation Proof that attests multiple keys per ETSI 119 472-3.

Batch Sign operations, have been removed to cleanup the codebase.

A new option called `proofs` has been added in `OpenId4VCIConfig` to allow a Wallet to configure the supported Proofs.

A Wallet can configure the following:
* Whether attestations that require no proofs are supported;
* Whether attestations that require proofs are supported; In this case a Wallet can also configure which JWS Algorithms are supported per Proof Type;

The new configuration option is taken into account just before generating the Proof for the Credential Request.
Before generating the Proof for the Credential Request, we check for the selected Credential Configuration:

* If it requires no proofs, whether Wallet supports attestations that require no proofs;
* If it requires proofs, whether the Credential Configuration contains any Wallet supported Proof Type. If no match is found, issuance fails;

pid-issuer test cases have been refactored to allow fetching a Key Attestation from a Wallet Provider.

Closes #509 
Addresses #520 (point 2)
Closes #508


dzarras:
@babisRoutis , @vafeini Could you provide some initial feedback?

babisRoutis:
@dzarras 

I would propose to firstly tackle #508

dzarras:
@babisRoutis, @vafeini Rebased on top of release/0.12.0, refined, and ready for review.

dzarras:
Changes from #517 have been included in this PR.

@babisRoutis, @vafeini Please review.

src/main/kotlin/eu/europa/ec/eudi/openid4vci/internal/http/CredentialIssuerMetadataJsonParser.kt
```diff
@@ -136,11 +138,11 @@ private object KeepKnownReusePolicies : JsonTransformingSerializer<CredentialReu
 @Serializable
 private data class ProofTypeSupportedMetaTO(
     @SerialName("proof_signing_alg_values_supported") val algorithms: List<String> = emptyList(),
-    @SerialName("key_attestations_required") val keyAttestationRequirement: KeyAttestationRequirementTO? = null,
+    @SerialName("key_attestations_required") val keyAttestationConstraints: KeyAttestationConstraintsTO? = null,
```
vafeini: I wouldn't rename it to KeyAttestationConstraints. It is what the issuer requires with reguards to key attestations passed from client, thus the name  KeyAttestationRequirement

dzarras: Name change has been reverted.

src/main/kotlin/eu/europa/ec/eudi/openid4vci/Config.kt
```diff
@@ -119,6 +121,7 @@ data class OpenId4VCIConfig(
     val clock: Clock = Clock.systemDefaultZone(),
     val issuerMetadataPolicy: IssuerMetadataPolicy = IssuerMetadataPolicy.IgnoreSigned,
     val supportedCredentialReusePolicies: CredentialReusePolicies? = null,
+    val proofs: ProofsConfig,
```
babisRoutis: Let's introduce a default value for this

src/main/kotlin/eu/europa/ec/eudi/openid4vci/Config.kt
```diff
@@ -134,6 +137,7 @@ data class OpenId4VCIConfig(
         clock: Clock = Clock.systemDefaultZone(),
         issuerMetadataPolicy: IssuerMetadataPolicy = IssuerMetadataPolicy.IgnoreSigned,
         supportedCredentialReusePolicies: CredentialReusePolicies? = null,
+        proofs: ProofsConfig,
```
babisRoutis: Let's use a default value

src/main/kotlin/eu/europa/ec/eudi/openid4vci/Config.kt
```diff
@@ -339,3 +344,33 @@ sealed interface IssuerMetadataPolicy {
...
 }
+
+/**
+ * Wallet supported proofs.
+ *
+ * @property isNoProofSupported whether the Wallet supports issuance of attestations that require no proofs
+ * @property jwtProof whether the Wallet supports JWT Proofs
+ * @property attestationProof whether the Wallet supports Attestation Proofs
+ */
+data class ProofsConfig(
```
babisRoutis, line 355: Consider adding a convinient factory method that would set both the `jwtProof` and the `attestationProof` with the same algorithms having the `isNoProofSupported` set to false

src/main/kotlin/eu/europa/ec/eudi/openid4vci/Config.kt
```diff
@@ -339,3 +344,33 @@ sealed interface IssuerMetadataPolicy {
...
+    data class SupportedJwtProof(val supportedAlgorithms: Set<JWSAlgorithm>)
+
+    /**
+     * Indicates support for Attestation Proofs.
+     *
+     * @property supportedAlgorithms the signing algorithms supported by the Wallet
+     */
+    data class SupportedAttestationProof(val supportedAlgorithms: Set<JWSAlgorithm>)
+
+    companion object
```
babisRoutis: Consider adding a default value

`isNoProofSupported` equal to false
`jwtproof` and `attestationProof` using the ES256

babisRoutis: @vafeini @dzarras what do you think?

An alternative could be  to allow exactly what ETSI 119 472-3 allows.
That is, true for the non-device-bound attestations and ES256, ES384, ES512 for the KA

babisRoutis (dismissed):
I think it is good.
Just left a comment wrt the use of a default value for the new config option

vafeini (dismissed):

src/main/kotlin/eu/europa/ec/eudi/openid4vci/Config.kt
vafeini: I agree, the whole PR is about going towards what specs and TS3 dictates

dzarras: ETSI 119 472-3 is basically what the extension property `ProofConfig.EC` in the test sources is: https://github.com/niscy-eudiw/eudi-lib-jvm-openid4vci-kt/blob/a758c6eacb57f12c771a9601af7d2f23d9e8482c/src/test/kotlin/eu/europa/ec/eudi/openid4vci/TestUtils.kt#L202

I could move it to the main sources, and rename it either to `Default` or `ETSI119472Part3`.

babisRoutis: please do that, use the `Default` name

cre8:
@dzarras @babisRoutis can you please point to the requirement that both approaches need to be supported by the issuer?

HAIP is forcing support for both by the **wallet**: https://openid.net/specs/openid4vc-high-assurance-interoperability-profile-1_0-final.html#name-key-attestation

There is no requirement for an issuer to support both, it is even optional. It will communicate via the issuer metadata which one the wallet should use.

babisRoutis:
> @dzarras @babisRoutis can you please point to the requirement that both approaches need to be supported by the issuer?
> 
> HAIP is forcing support for both by the **wallet**: https://openid.net/specs/openid4vc-high-assurance-interoperability-profile-1_0-final.html#name-key-attestation
> 
> There is no requirement for an issuer to support both, and if so not to support both. It will communicate via the issuer metadata which one the wallet should use.

Dear @cre8 

There is no requirement for the issuer to support both proof of type `jwt` with key attestation and proof of type `attestation`, at the same time.
 According, HAIP v1, TS3 and ETSI TS 119 472-3  at least one option must be used.. 

Yet I don't understand the point of your question, in the context of this closed PR.
Can you please elaborate?

cre8:
> > @dzarras @babisRoutis can you please point to the requirement that both approaches need to be supported by the issuer?
> > HAIP is forcing support for both by the **wallet**: https://openid.net/specs/openid4vc-high-assurance-interoperability-profile-1_0-final.html#name-key-attestation
> > There is no requirement for an issuer to support both, and if so not to support both. It will communicate via the issuer metadata which one the wallet should use.
> 
> Dear @cre8
> 
> There is no requirement for the issuer to support both proof of type `jwt` with key attestation and proof of type `attestation`, at the same time. According, HAIP v1, TS3 and ETSI TS 119 472-3 at least one option must be used..
> 
> Yet I don't understand the point of your question, in the context of this closed PR. Can you please elaborate?

Hi @babisRoutis ,

in the PR description you wrote: Credential Issuers are now required to support **both** JWT Proofs with Key Attestation and Attestation Proofs per TS3/ETSI 119 472-3.

Here in the credential configuration you either say none is there or both has to be there: https://github.com/eu-digital-identity-wallet/eudi-lib-jvm-openid4vci-kt/blob/main/src/main/kotlin/eu/europa/ec/eudi/openid4vci/CredentialConfiguration.kt#L148

This is causing some breaks on my side since the current issuers are only doing JWT.

babisRoutis:
Dear @cre8 

The requirement on the code is too strict, and this `&&` definitely should be replaced by `||`. 
You can track the fix https://github.com/eu-digital-identity-wallet/eudi-lib-jvm-openid4vci-kt/issues/566#issue-4991000845


Apologies for causing issues, and thank you for reporting this.

paulbastian:
I would go even further and says that it's not the wallets job to check what the issuer requires. It's the issuers job to ask for key attestations for eIDAS-compatible flows. We would like the wallet to support issuance without key attestations!

cre8:
> I would go even further and says that it's not the wallets job to check what the issuer requires. It's the issuers job to ask for key attestations for eIDAS-compatible flows. We would like the wallet to support issuance without key attestations!

I think this is already possible, it is right now only checking for both when it is not empty (but I am not a kotlin expert):
https://github.com/eu-digital-identity-wallet/eudi-lib-jvm-openid4vci-kt/blob/main/src/main/kotlin/eu/europa/ec/eudi/openid4vci/CredentialConfiguration.kt#L145

Sakurann:
I am afraid current code does not allow issuers to issue EAAs without key binding if issuer supports proof types. that needs to be fixed too. made a comment in https://github.com/eu-digital-identity-wallet/eudi-lib-jvm-openid4vci-kt/issues/566#issuecomment-5096715299

babisRoutis:
> I would go even further and says that it's not the wallets job to check what the issuer requires. It's the issuers job to ask for key attestations for eIDAS-compatible flows. We would like the wallet to support issuance without key attestations!

I totally disagree on that.

babisRoutis:
@cre8 Hi Mirko

FYI
1 -> https://github.com/eu-digital-identity-wallet/eudi-lib-jvm-openid4vci-kt/issues/566#issuecomment-5100864056
2 -> https://github.com/eu-digital-identity-wallet/eudi-lib-jvm-openid4vci-kt/releases/tag/v0.12.2

PR #517, opened 2026-04-15
Add an option for a Wallet to configure supported Proof types

### LINKED ISSUE #508: Introduce a configuration option to enforce the use of key_attestation

dzarras:
This PR add a new option called `proofs` in `OpenId4VCIConfig` as a means for a Wallet to configure the supported Attestation/Proofs.

Current a Wallet can configuration the following:
1. Whether non-device-bound attestations are supported;
2. Whether device-bound attestations are supported; For device-bound attestations a Wallet can also:
    1. Configure which JWS Algorithms are supported for the Proofs;
    2. Configure which Proof Types are supports; Currently the supports Proof Types are:
        1. JWT Proofs without Key Attestation;
        2. JWT Proofs with Key Attestation;
        3. Attestation Proofs;

The new configuration option is taken into account just before generating the Proofs for the Credential Request.
Before generating the Proofs for the Credential Request, we check for the selected Credential Configuration:
* If it's a non-device-bound attestation, whether Wallet supports non-device-bound attestations;
* If it's a device-bound attestation, whether the Credential Configuration contains any Wallet supported Proof Type. If no match is found, issuance fails;

Closes #508

### DIFF
--- a/src/main/kotlin/eu/europa/ec/eudi/openid4vci/Config.kt
+++ b/src/main/kotlin/eu/europa/ec/eudi/openid4vci/Config.kt
@@ -104,8 +104,7 @@ data class OpenId4VCIConfig(
 ) {
 
     /**
-     * Creates a new [OpenId4VCIConfig] instance for a Wallet that uses [a Public OAuth 2.0 Client][ClientAuthentication.None], and all
-     * [Proof Types][ProofsConfig].
+     * Creates a new [OpenId4VCIConfig] instance for a Wallet that uses [a Public OAuth 2.0 Client][ClientAuthentication.None].
      */
     constructor(
         clientId: ClientId,
@@ -117,6 +116,7 @@ data class OpenId4VCIConfig(
         clock: Clock = Clock.systemDefaultZone(),
         issuerMetadataPolicy: IssuerMetadataPolicy = IssuerMetadataPolicy.IgnoreSigned,
         supportedCredentialReusePolicies: CredentialReusePolicies? = null,
+        proofs: ProofsConfig,
     ) : this(
         ClientAuthentication.None(clientId, dPoPUsage),
         authFlowRedirectionURI,
@@ -126,7 +126,7 @@ data class OpenId4VCIConfig(
         clock,
         issuerMetadataPolicy,
         supportedCredentialReusePolicies,
-        ProofsConfig.All,
+        proofs,
     )
 }
 
@@ -338,18 +338,6 @@ data class ProofsConfig(
     val supportsDeviceBound: Boolean
         get() = null != deviceBound
 
-    companion object {
-        val All: ProofsConfig = ProofsConfig(
-            supportsNonDeviceBound = true,
-            deviceBound = DeviceBound(
-                null,
-                DeviceBound.Proof.JwtProofWithoutKeyAttestation,
-                DeviceBound.Proof.JwtProofWithKeyAttestation,
-                DeviceBound.Proof.AttestationProof,
-            ),
-        )
-    }
-
     /**
      * The Proofs a Wallet supports for device-bound attestations.
      *
@@ -357,13 +345,11 @@ data class ProofsConfig(
      * @property proofs The Proofs the Wallet supports for device-bound attestations.
      */
     data class DeviceBound(
-        val algorithms: Set<JWSAlgorithm>?,
+        val algorithms: Set<JWSAlgorithm>,
         val proofs: Set<Proof>,
     ) {
-        constructor(algorithms: Set<JWSAlgorithm>?, proof: Proof, vararg proofs: Proof) : this(algorithms, setOf(proof, *proofs))
-
         init {
-            require(null == algorithms || algorithms.isNotEmpty()) { "At least one algorithm must be supported" }
+            require(algorithms.isNotEmpty()) { "At least one algorithm must be supported" }
             require(proofs.isNotEmpty()) { "At least one proof must be supported" }
         }
 
@@ -373,4 +359,6 @@ data class ProofsConfig(
             data object AttestationProof : Proof
         }
     }
+
+    companion object
 }
--- a/src/main/kotlin/eu/europa/ec/eudi/openid4vci/Config.kt
+++ b/src/main/kotlin/eu/europa/ec/eudi/openid4vci/Config.kt
@@ -18,6 +18,7 @@ package eu.europa.ec.eudi.openid4vci
 import com.nimbusds.jose.CompressionAlgorithm
 import com.nimbusds.jose.EncryptionMethod
 import com.nimbusds.jose.JWEAlgorithm
+import com.nimbusds.jose.JWSAlgorithm
 import com.nimbusds.jose.crypto.ECDHEncrypter
 import com.nimbusds.jose.crypto.impl.ContentCryptoProvider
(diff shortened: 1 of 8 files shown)


dzarras:
@babisRoutis, @vafeini PR now targets release/0.12.0 and is ready for review.

src/main/kotlin/eu/europa/ec/eudi/openid4vci/Config.kt
```diff
@@ -375,3 +384,53 @@ sealed interface IssuerMetadataPolicy {
...
+    val supportsDeviceBound: Boolean
+        get() = null != deviceBound
+
+    companion object {
+        val All: ProofsConfig = ProofsConfig(
+            supportsNonDeviceBound = true,
+            deviceBound = DeviceBound(null, DeviceBound.Proof.entries.toSet()),
+        )
+
+        val None: ProofsConfig = ProofsConfig(
```
babisRoutis: That's not useful. Let's remove it please

src/main/kotlin/eu/europa/ec/eudi/openid4vci/Config.kt
```diff
@@ -375,3 +384,53 @@ sealed interface IssuerMetadataPolicy {
...
+    }
+
+    /**
+     * The Proofs a Wallet supports for device-bound attestations.
+     *
+     * @property algorithms The JWS Algorithms the Wallet supports for device-bound attestations. Set to *null* if the Wallet supports all algorithms.
+     * @property proofs The Proofs the Wallet supports for device-bound attestations.
+     */
+    data class DeviceBound(
+        val algorithms: Set<JWSAlgorithm>?,
```
babisRoutis: Why is this nullable?

src/main/kotlin/eu/europa/ec/eudi/openid4vci/Config.kt
```diff
@@ -375,3 +384,53 @@ sealed interface IssuerMetadataPolicy {
...
+        val proofs: Set<Proof>,
+    ) {
+        init {
+            require(null == algorithms || algorithms.isNotEmpty()) { "At least one algorithm must be supported" }
+            require(proofs.isNotEmpty()) { "At least one proof must be supported" }
+        }
+
+        enum class Proof {
```
babisRoutis: Please consider making this a sealed hierarchy.
In the future, we can add more requirements/features for each case

src/main/kotlin/eu/europa/ec/eudi/openid4vci/Config.kt
dzarras: The idea was that a Wallet can optionally limit the supported JWS Algorithms. 
On seconds thoughts though, it doesn't really make sense, a Wallet will always have a specific set of supported JWS Algorithms.

I'll update it to be non-nullable.

src/main/kotlin/eu/europa/ec/eudi/openid4vci/Config.kt
```diff
@@ -375,3 +384,53 @@ sealed interface IssuerMetadataPolicy {
...
+/**
+ * Wallet supported Proofs.
+ *
+ * A Wallet has to configure whether it supports non-device-bound attestations, as well as the Proofs supported for device-bound attestations.
+ *
+ * @property supportsDeviceBound Whether the Wallet supports non-device-bound attestations.
+ * @property deviceBound Whether the Wallet supports device-bound attestations, and the Proofs it supports.
+ */
+data class ProofsConfig(
+    val supportsNonDeviceBound: Boolean,
```
babisRoutis, line 335: Can you please rename it to `isNoProofSupported`?

The non-device-bound term is not recognizable immediately in a OpenId4VCI context.

src/main/kotlin/eu/europa/ec/eudi/openid4vci/Config.kt
```diff
@@ -375,3 +384,53 @@ sealed interface IssuerMetadataPolicy {
...
+
+    /**
+     * The Proofs a Wallet supports for device-bound attestations.
+     *
+     * @property algorithms The JWS Algorithms the Wallet supports for device-bound attestations. Set to *null* if the Wallet supports all algorithms.
+     * @property proofs The Proofs the Wallet supports for device-bound attestations.
+     */
+    data class DeviceBound(
+        val algorithms: Set<JWSAlgorithm>?,
+        val proofs: Set<Proof>,
```
babisRoutis, line 349: I don't like the `Set` approach, but it works

src/main/kotlin/eu/europa/ec/eudi/openid4vci/Config.kt
dzarras: Removed both All and None.

src/main/kotlin/eu/europa/ec/eudi/openid4vci/Config.kt
dzarras: Done. `algorithms` is non-null and must be explicitly defined by the Wallet.

src/main/kotlin/eu/europa/ec/eudi/openid4vci/Config.kt
dzarras: Done. `Proof` is now a `sealed interface` and `enum` constants have been converted to `data objects`

src/main/kotlin/eu/europa/ec/eudi/openid4vci/Config.kt
dzarras: Device-bound as a term is used neither in OpenID4VCI nor HAIP.

If we rename the property `supportsNonDeviceBound` we also have to rename the property `deviceBound` and the data class `DeviceBound` as well.

Any ideas/suggestions for the later?

src/main/kotlin/eu/europa/ec/eudi/openid4vci/Config.kt
babisRoutis: @dzarras Can you please consider https://github.com/eu-digital-identity-wallet/eudi-lib-jvm-openid4vci-kt/issues/508#issuecomment-4206521475

Let's discuss this offline

dzarras:
@babisRoutis, @vafeini Rebased against release/0.12.0. Please check again.

dzarras:
Rebased on top of release/0.12.0

dzarras:
Superseded by #512

PR #523, opened 2026-04-21
Expose a way for clients to manually refresh access tokens

### LINKED ISSUE #522: Expose a way to clients to manually refresh access tokens
Currently the library transparently handles refreshing expired access tokens when querying for deferred credentials.

It would be helpful though to expose this functionality to clients and allow them to manually refresh access token when deemed necessary.

dzarras:
This PR introduces `RefreshAccessToken` as a new capability to `Issuer` and `DeferredIssuer`. 

It allows clients to manually run a `refresh_token` grant if needed.

This functionality already existed in the library and was transparently handled by the library when querying for deferred credentials. 
This PR effectively makes it public.

Closes #522

### DIFF
--- a/src/main/kotlin/eu/europa/ec/eudi/openid4vci/DeferredIssuer.kt
+++ b/src/main/kotlin/eu/europa/ec/eudi/openid4vci/DeferredIssuer.kt
@@ -18,7 +18,7 @@ package eu.europa.ec.eudi.openid4vci
 import com.nimbusds.jose.CompressionAlgorithm
 import com.nimbusds.jose.EncryptionMethod
 import com.nimbusds.jose.jwk.JWK
-import eu.europa.ec.eudi.openid4vci.internal.RefreshAccessToken
+import eu.europa.ec.eudi.openid4vci.internal.RefreshAccessTokenImpl
 import eu.europa.ec.eudi.openid4vci.internal.http.DeferredEndPointClient
 import eu.europa.ec.eudi.openid4vci.internal.http.TokenEndpointClient
 import io.ktor.client.*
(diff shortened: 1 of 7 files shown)


src/main/kotlin/eu/europa/ec/eudi/openid4vci/RefreshAccessToken.kt
```diff
@@ -0,0 +1,33 @@
...
+    /**
+     * Performs a `refresh_token` grant against the Token Endpoint of the Authorization Server to fetch a new Access Token, and Refresh Token.
+     */
+    suspend fun AuthorizedRequest.refresh(): Result<AuthorizedRequest>
+
+    /**
+     * Performs a `refresh_token` grant against the Token Endpoint of the Authorization Server to fetch a new Access Token, and Refresh Token,
+     * in case the current Access Token is expired.
+     */
+    suspend fun AuthorizedRequest.refreshIfNeeded(): Result<AuthorizedRequest>
```
vafeini: Do we need 2 methods? Isn't refreshIfNeeded the default case?

src/main/kotlin/eu/europa/ec/eudi/openid4vci/DeferredIssuer.kt
```diff
@@ -183,7 +183,7 @@ interface DeferredIssuer : QueryForDeferredCredential {
                 httpClient,
             )
 
-            val refreshAccessToken = RefreshAccessToken(config.clock, tokenEndpointClient)
+            val refreshAccessToken = RefreshAccessTokenImpl(config.clock, tokenEndpointClient)
```
babisRoutis: I have the impression that the presence of TokenEndPoint is not enough to execute refresh token grant.
Can you please check ?
I think that in addition Authorization server must advertise `refresh_token`

vafeini (approved):

src/main/kotlin/eu/europa/ec/eudi/openid4vci/QueryForDeferredCredential.kt
```diff
@@ -88,10 +90,13 @@ fun interface QueryForDeferredCredential {
                 refreshed.withResourceServerDpopNonce(newResourceServerDpopNonce) to outcome
             }
 
-            private suspend fun refreshIfNeeded(authorizedRequest: AuthorizedRequest): AuthorizedRequest =
-                with(refreshAccessToken) {
-                    authorizedRequest.refreshIfNeeded().getOrThrow()
+            private suspend fun refreshIfNeeded(authorizedRequest: AuthorizedRequest): AuthorizedRequest {
```
vafeini, line 93: I think it would make sense to have this behavior as the default behavior of RefreshAccessTokenImpl.refresh()

PR #526, opened 2026-04-27
Narrow the encryption algorithms for credential request/response encryption

### LINKED ISSUE #510: Narrow the encryption algorithms for credential request/response encryption
Currently, the library allows the caller to define which algorithms are allowed for credential request/response encryption, via classes 
- `RsaConfig` and
- `EcConfig` 

Especially the former (`RsaConfig`) 
- Checks that the user provided algorithms are in `RSAEncrypter.SUPPORTED_ALGORITHMS` (provided by nimbus) and
- Defaults `supportedJWEAlgorithms` to the above list.

In addition, `EncryptionSupportConfig` has a similar problem given that it is uses `ContentCryptoProvider.SUPPORTED_ENCRYPTION_METHODS` which includes deprecated methods.

atsant:
closes #510

### DIFF
--- a/src/main/kotlin/eu/europa/ec/eudi/openid4vci/Config.kt
+++ b/src/main/kotlin/eu/europa/ec/eudi/openid4vci/Config.kt
@@ -231,11 +231,11 @@ data class EncryptionSupportConfig(
     val credentialResponseEncryptionPolicy: CredentialResponseEncryptionPolicy,
     val ecConfig: EcConfig?,
     val rsaConfig: RsaConfig?,
-    val supportedEncryptionMethods: List<EncryptionMethod> = allowedEncryptionMethods,
+    val supportedEncryptionMethods: List<EncryptionMethod> = SUPPORTED_ENCRYPTION_ALGORITHMS.toList(),
 ) {
     init {
         require(supportedEncryptionMethods.isNotEmpty()) { "At least one encryption method must be provided" }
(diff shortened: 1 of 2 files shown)


src/main/kotlin/eu/europa/ec/eudi/openid4vci/Config.kt
```diff
@@ -278,6 +283,11 @@ data class RsaConfig(
     }
 }
 
+private val allowedRSAAlgorithms: List<JWEAlgorithm> get() = RSAEncrypter.SUPPORTED_ALGORITHMS.allowedAlgorithms()
+private fun Set<JWEAlgorithm>.allowedAlgorithms(): List<JWEAlgorithm> = this.filterNot {
+    it in setOf(JWEAlgorithm.RSA1_5, JWEAlgorithm.RSA_OAEP)
+}
```
atsant: With this way we can (later if needed) to deprecate aglorithms for `EcConfig` too. in a single place.

dzarras (dismissed):

dzarras:
@atsant Please rebase against `main`

dzarras:
Unrelated to the PR content. 

Didn't a rebase used to make an approval stale?

atsant:
@dzarras 
I believe it is making it stale only if the rebase had conflicts and changes had to be made in the reviewed code.
Since the pr scopes has not been affected by the rebase it might think that review still stands.

Not so sure about it though.

dzarras (approved):

PR #528, opened 2026-04-30
Refactor `ClientAttestation JWT` provisioning when `ClientAuthentication.AttestationBased` is used

### LINKED ISSUE #525: Improve Attestation-Based Client Authentication
## Problem

Currently, the wallet declares that it supports attestation-based client authentication using the following option

```kotlin
data class AttestationBased(
        val attestationJWT: ClientAttestationJWT,
        val popJwtSpec: ClientAttestationPoPJWTSpec,
) : ClientAuthentication
```

In addition, it has to provide to VCI configuration the following:

```kotlin
val clientAttestationPoPBuilder: ClientAttestationPoPBuilder
```

Given that both of these attributes need to be provided in order to obtain an instance of `Issuer` interface and proceed with resolution of a credential offer, there is the following problem: `ClientAttestationJWT` needs to be provisioned during negotiation of the wallet with the credential issuer, but the library expects this to already be provisioned.

## Solution

Introduce a new interface that would deffer provisioning of `ClientAttestationJWT` after the resolution of the credential offer and before proceeding with the authorization and the issuance.

This would require the introduction of an interface like bellow

```kotlin
interface ProvisionClientAttestation {
    val algorithm: JwsAlgorithm

    suspend fun provision(authorizationServer: URL): Provisioned

    data class Provisioned(
        val attestation: ClientAttestationJWT,
        val createPoPJWTResource: ClientAttestationPoPJWTSpec,
    )
}
```

The interface defines a single function that the wallet must implement. 
Provisions both the `ClientAttestationJWT` and the ability to sign a PoP JWT in a single step, in the context of an authorization server 

Then the definition of AttestationBased (Client) can be refactored to 

```kotlin
data class AttestationBased(
        override val id: ClientId,
        val provisionClientAttestation : ProvisionClientAttestation,
) : ClientAuthentication
```

dzarras:
Closes #525


dzarras:
@babisRoutis, @vafeini Could you take an quick look and provide initial comments? 

This PR currently targets `main` but introduces breaking changes. We must consider which branch this refactor should land in. Currently `release/0.11.0` contains changes related to TS3. Would it make sense to target `release/0.11.0`?

src/main/kotlin/eu/europa/ec/eudi/openid4vci/ClientAttestation.kt
```diff
@@ -128,8 +128,9 @@ fun interface ClientAttestationPoPBuilder {
      *
      * @return the PoP JWT
      */
-    suspend fun ClientAuthentication.AttestationBased.attestationPoPJWT(
+    suspend fun ClientAttestationPoPJWTSpec.attestationPoPJWT(
         clock: Clock,
+        clientId: ClientId,
```
babisRoutis: `client_id` is not mandatory in ABCA. It is totally valid to place request to PAR and/or Token endpoint without including an explicit `client_id` parameter. If it is included, it has to match the contents of the jwt(s)

No action for now. Unfortunatelly, keycloak requires `client_id`

src/main/kotlin/eu/europa/ec/eudi/openid4vci/Config.kt
```diff
@@ -47,16 +63,9 @@ sealed interface ClientAuthentication : java.io.Serializable {
...
-    ) : ClientAuthentication {
-        init {
-            require(attestationJWT.clientId.isNotBlank())
-            require(!attestationJWT.publicKey.isPrivate) { "InstanceKey should be public" }
-        }
-
-        override val id: ClientId get() = attestationJWT.clientId
-    }
+        override val id: ClientId,
+        val provisionClientAttestation: ProvisionClientAttestation,
```
babisRoutis, line 65: The config contains a closely related attribute 
`clientAttestationPoPBuilder`

Perhaps this could be removed from vci config and added to `ProvisionClientAttestation.Provisioned`.

Alternatively, if it could be moved to `AttestationBased`

dzarras: Moved `ClientAttestationPoPBuilder` in `ProvisionClientAttestation.Provisioned`.

dzarras:
@babisRoutis, @vafeini PR now targets release/0.12.0 and is ready for review.

babisRoutis:
@vafeini , @dzarras 

I have a specific change in mind, but I am not sure if we should implemented to this PR

The idea is to drop totally support for other client authentication methods besides ABCA.
Reason for this drastic change is the following:

Normally, DPoP is an orthogonal feature with regards to an Client Authentication Method. That is, they are unrelated.
This "normal" case is reflected to the our VCI Configuration.

On the other hand, we have some strong "signals" that this won't be the case for EUDIW.
- For starters there would be TS3 v1.5.1, and
- There would a next ABCA version 

Both of these, contain the same requirement that puts a strong relation between ABCA and DPoP:
- The Client Instance Key is allocated per Authorization Server (protecting a Credential Issuer)
- The Client Instance Key is included to the Client Attestation JWT (or WIA), under `cnf.jwk`
- The private material of Client Instance Key, signs the Client Attestation PoP JWT
- The same is true for the DPoP.

This means, that DPoP configuration and ABCA can no longer remain independent of each other, to the configuration of our library.

Alternatively we may seek a solution in between. Effectively I see two options

Option 1: 
- Support exclusively ABCA
- Unify Client Attestation PoP and DPoP options (under ABCA)

Option 2:
- Unify Client Attestation PoP and DPoP options (under ABCA)
- Keep the top-level DPoP options
- If ABCA, require empty top-level DPoP options (because ABCA will be used)
- If not ABCA, make the top-level options relevant

Let's discuss it please

dzarras:
> @vafeini , @dzarras
> 
> I have a specific change in mind, but I am not sure if we should implemented to this PR
> 
> The idea is to drop totally support for other client authentication methods besides ABCA. Reason for this drastic change is the following:
> 
> Normally, DPoP is an orthogonal feature with regards to an Client Authentication Method. That is, they are unrelated. This "normal" case is reflected to the our VCI Configuration.
> 
> On the other hand, we have some strong "signals" that this won't be the case for EUDIW.
> 
> * For starters there would be TS3 v1.5.1, and
> * There would a next ABCA version
> 
> Both of these, contain the same requirement that puts a strong relation between ABCA and DPoP:
> 
> * The Client Instance Key is allocated per Authorization Server (protecting a Credential Issuer)
> * The Client Instance Key is included to the Client Attestation JWT (or WIA), under `cnf.jwk`
> * The private material of Client Instance Key, signs the Client Attestation PoP JWT
> * The same is true for the DPoP.
> 
> This means, that DPoP configuration and ABCA can no longer remain independent of each other, to the configuration of our library.
> 
> Alternatively we may seek a solution in between. Effectively I see two options
> 
> Option 1:
> 
> * Support exclusively ABCA
> * Unify Client Attestation PoP and DPoP options (under ABCA)
> 
> Option 2:
> 
> * Unify Client Attestation PoP and DPoP options (under ABCA)
> * Keep the top-level DPoP options
> * If ABCA, require empty top-level DPoP options (because ABCA will be used)
> * If not ABCA, make the top-level options relevant
> 
> Let's discuss it please

I would tackle this in a separate PR to reduce the churn.

I would go with Option 1.

dzarras:
@babisRoutis, @vafeini Additional changes have been incorporated to the PR. Please review.

src/main/kotlin/eu/europa/ec/eudi/openid4vci/internal/http/AuthorizationEndpointClient.kt
```diff
@@ -274,12 +276,16 @@ internal class AuthorizationEndpointClient(
                 dPoPJwtFactory?.createDPoPJwt(Htm.POST, url, null, existingDpopNonce)
                     ?.getOrThrow()?.serialize()
 
-            val abcaChallenge = when (config.clientAuthentication) {
-                is ClientAuthentication.AttestationBased ->
-                    existingAbcaChallenge ?: challengeEndpointClient?.getChallenge()?.getOrThrow()
-                else -> null
-            }
-            val clientAttestation = config.generateClientAttestationIfNeeded(URI(authorizationIssuer).toURL(), abcaChallenge)
+            val abcaChallenge =
```
babisRoutis: I am not sure about this with regards to DPoP Nonce

When calling the challenge endpoint, it is perfectly valid to obtain with the response a DPoP Nonce.

Currently, we do use DPoP Nonce (`existingDpopNonce`), if any, but there is the case that DPoPNonce changes because of invoking the challenge endpoint

dzarras: @babisRoutis , @vafeini Pushed an update to also consider any `DPoP-Nonce` returned by the Challenge Endpoint. Please check again.

src/main/kotlin/eu/europa/ec/eudi/openid4vci/internal/http/AuthorizationEndpointClient.kt
```diff
@@ -264,35 +266,60 @@ internal class AuthorizationEndpointClient(
             }
         }
 
+        suspend fun abcaChallengeAndDPoPNonce(
+            existingAbcaChallenge: Nonce?,
+            existingDpopNonce: Nonce?,
+        ): Pair<Nonce?, Nonce?> {
+            if (null == provisionedClientAttestation) {
+                check(null == existingAbcaChallenge) { "can't have abca challenge with non-attested client" }
```
babisRoutis: Do we really need this check?

dzarras: Not really. Just a sanity check that can be omitted.

src/main/kotlin/eu/europa/ec/eudi/openid4vci/internal/http/TokenEndpointClient.kt
```diff
@@ -198,34 +199,61 @@ internal class TokenEndpointClient(
...
-        val params = TokenEndpointForm.refreshAccessToken(clientAuthentication.id, refreshToken)
+        val params = TokenEndpointForm.refreshAccessToken(clientId, refreshToken)
         placeTokenRequest(params, dpopNonce)
     }
 
     private suspend fun placeTokenRequest(
         params: Map<String, String>,
         dpopNonce: Nonce?,
     ): Pair<TokenResponse, Nonce?> {
+        suspend fun abcaChallengeAndDPoPNonce(
```
babisRoutis: Would it make sense to expose this as an internal top-level function, 

Using the existing two parameters and having two context parameters `provisionedClientAttestation` and `challengeEndpointClient'

For instance, 

```kotlin

context(provisionedClientAttestation: ProvisionClientAttestation.Provisioned?, challengeEndpointClient: ChallengeEndpointClient?)
internal suspend fun  abcaChallengeAndDPoPNonce(
    existingAbcaChallenge: Nonce?,
    existingDpopNonce: Nonce?
): Pair<Nonce?, Nonce?>

```

On the call site 

```kotlin

context(provisionedClientAttestation, challengeEndpointClient) {
    abcaChallengeAndDPoPNonce(existingAbcaChallenge, existingDpopNonce)
}

```

dzarras: Yes it does. I didn't do it this way because we don't have Context Parameters enabled.
Should I enable them, and update the function as described?

dzarras: Tried enabling context parameters and updating the method as discussed.
Code-wise it works.
Unfortunately, we use an old version of ktlint, and it doesn't support context parameters.
I don't want to update ktlint in this PR to avoid the churn.

I propose to :
1. Proceed with this PR as is;
2. Update ktlint and reformat the code in a separate PR;
3. Enable context parameters and revisit this method in a new PR;

babisRoutis: @dzarras 

Let's take the simpler approach

1. Either declare the context parameters as simple parameters, or
2. Declare simple utility class that accepts in its constructor the two context parameters

I would prefer the 2nd, but I leaving this on you.

babisRoutis: My comment was about avoiding duplicate code, not about using context parameters.

src/main/kotlin/eu/europa/ec/eudi/openid4vci/internal/http/AuthorizationEndpointClient.kt
babisRoutis: Then let's remove it.

dzarras: Done

src/main/kotlin/eu/europa/ec/eudi/openid4vci/internal/http/TokenEndpointClient.kt
dzarras: Introduced a new internal utility class.

vafeini (approved):

babisRoutis (approved):

PR #500, opened 2026-03-26
Align ClientAttestation JWT claims with Wallet Instance Attestation

dzarras:
Closes #496

PR #531, opened 2026-05-07
Update KeyAttestation according to TS3 v1.5 and IA

### LINKED ISSUE #521: Key attestation validation updates to align with TS3 v1.5
1. Validate structure of ΚΑ to be aligned with [TS3 data model](https://github.com/eu-digital-identity-wallet/eudi-doc-standards-and-technical-specifications/blob/main/docs/technical-specifications/ts3-wallet-unit-attestation.md#232-content-of-key-attestation-ka) 
2. Validate that ΚΑ is singed and one of the following algorithms were used: ES256, ES384, ES512
3. For a JWT proof type with ΚΑ in header
    - JWT must be signed by the first key included in the `attested_keys` array of ΚΑ
    - JWT header should contain claim `kid` with value 0.

dzarras:
~~**Builds on top of and requires #500 to be merged first.**~~

Closes #521


dzarras:
Rebased and ready for review.

src/main/kotlin/eu/europa/ec/eudi/openid4vci/KeyAttestationJWT.kt
```diff
@@ -15,60 +15,92 @@
...
+            val claimsSet = jwt.ensureValidClaimsSet<KeyAttestationJWTClaims>()
+            return KeyAttestationJWT(jwt.serialize(), jwt.header, claimsSet)
+        }
     }
+}
 
-    companion object {
+fun KeyAttestationJWT.serialize(): String = jwt
+
+typealias JWKList = List<
```
babisRoutis: I find the name `JWKList` confusing (being very close to JWK set)

Consider renaming to `AttestedKeys`

src/main/kotlin/eu/europa/ec/eudi/openid4vci/KeyAttestationJWT.kt
```diff
@@ -15,60 +15,92 @@
...
+    @Required @SerialName(OpenId4VCISpec.USER_AUTHENTICATION) val userAuthentication: List<AttackPotentialResistance>,
+    @Required @SerialName(OpenId4VCISpec.CERTIFICATION) @Serializable(with = URLSerializer::class) val certification: URL,
+    @SerialName(OpenId4VCISpec.NONCE) val nonce: Nonce? = null,
+    @SerialName(TokenStatusListSpec.STATUS) val status: StatusClaim? = null,
+    @Required @SerialName(TS3.KEY_STORAGE_STATUS) val keyStorageStatus: KeyStorageStatus,
+) {
+    init {
+        require(attestedKeys.isNotEmpty()) { "attestedKeys must not be empty" }
+        require(attestedKeys.none { it.isPrivate }) { "attestedKeys must all be public" }
+        require(1 == keyStorage.size && AttackPotentialResistance.Iso18045High == keyStorage.first()) {
```
babisRoutis: I am not sure if this rule is correct.

I assume that this rule comes from TS3 

> For a KA about a WSCD, the key_storage and user_authentication attributes shall be iso_18045_high as the WSCD by definition must ensure LoA High

I am not sure about the size on the restriction of the list (to be exactly 1)



I would recommend:

Add to the companion a suitable function that enforces the rule and documents that it is a TS3-specified rule.
Then use it to the init { }.

Do the same for the following similar rule about the `user_authentication`

Finally, consider adding a factory method that omits the keyStorage and userAthentication parameters, calling the default constructor with the suitable values.

src/main/kotlin/eu/europa/ec/eudi/openid4vci/StatusClaim.kt
```diff
@@ -0,0 +1,31 @@
...
+
+@Serializable
+data class StatusClaim(
+    @Required @SerialName(TokenStatusListSpec.STATUS_LIST) val statusList: StatusListTokenClaim,
+)
+
+@Serializable
+data class StatusListTokenClaim(
+    @Required @SerialName(TokenStatusListSpec.INDEX) val index: UInt,
+    @Required @SerialName(TokenStatusListSpec.URI) val uri: NonBlankString,
```
babisRoutis: Why is this a `NonBlankString` and not a `URL`

Library provides already `URLSerializer`

src/main/kotlin/eu/europa/ec/eudi/openid4vci/ClientAttestation.kt
```diff
@@ -47,6 +47,7 @@ typealias ClientAttestation = Pair<ClientAttestationJWT, ClientAttestationPoPJWT
  * be used by the instance for client authentication
  */
 @ConsistentCopyVisibility
+@Serializable(with = ClientAttestationJWTSerializer::class)
```
babisRoutis: Is there a need to "Serialize" `ClientAttestationJWT` ?

If not, let's remove the annotation and the custom serializer

src/main/kotlin/eu/europa/ec/eudi/openid4vci/ClientAttestation.kt
```diff
@@ -202,14 +194,14 @@ internal fun SignedJWT.ensureSignedOrVerified() {
     }
 }
 
-private fun SignedJWT.ensureSignedWithAllowedAlgorithm() {
-    require(header.algorithm in TS3.WALLET_INSTANCE_ATTESTATION_ALLOWED_SIGNATURE_ALGORITHMS) {
-        "Invalid Attestation JWT. Signature algorithm must be one of ${TS3.WALLET_INSTANCE_ATTESTATION_ALLOWED_SIGNATURE_ALGORITHMS}"
+internal fun SignedJWT.ensureSignedWithAllowedAlgorithm(allowedAlgorithms: Set<JWSAlgorithm>) {
```
babisRoutis: Since the methods are generic and given that are marked as internal,
consider moving them to internal package to a dedicated file

src/main/kotlin/eu/europa/ec/eudi/openid4vci/Issuance.kt
```diff
@@ -146,15 +146,18 @@ sealed interface ProofsSpecification {
 
         data class WithKeyAttestation(
             val proofSignerProvider: suspend (Nonce?) -> Signer<KeyAttestationJWT>,
-            val keyIndex: Int,
-        ) : JwtProofs
+        ) : JwtProofs {
+            val keyIndex: Int = 0
```
babisRoutis: I think that it doesn't make sense to have this constant.

Either remove the `keyIndex` and take for granted that it is zero. 
or leave it as it was, and check the invariant

src/main/kotlin/eu/europa/ec/eudi/openid4vci/internal/RequestIssuanceImpl.kt
```diff
@@ -408,20 +412,20 @@ internal class RequestIssuanceImpl(
...
-                val verifier: JWSVerifier? = when (jwk) {
-                    is RSAKey -> RSASSAVerifier(jwk)
-                    is ECKey -> ECDSAVerifier(jwk)
-                    else -> null
-                }
-                verifier != null && jwtProof.verify(verifier)
-            } catch (_: Exception) {
-                false
-            }
+        val verifier: JWSVerifier = when (val jwk = attestedKeys.first()) {
```
babisRoutis: Is this aligned with the requirement of using EC algs ?

I guess no

babisRoutis (changes requested):

src/main/kotlin/eu/europa/ec/eudi/openid4vci/StatusClaim.kt
```diff
@@ -0,0 +1,31 @@
...
+ * limitations under the License.
+ */
+package eu.europa.ec.eudi.openid4vci
+
+import kotlinx.serialization.Required
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+
+@Serializable
+data class StatusClaim(
```
vafeini: Do we need it as a separate file? Why not include it in Types.kt?

src/main/kotlin/eu/europa/ec/eudi/openid4vci/KeyAttestationJWT.kt
```diff
@@ -15,60 +15,92 @@
...
+            jwt.ensureSignedOrVerified()
+            jwt.ensureSignedWithAllowedAlgorithm(TS3.WALLET_INSTANCE_ATTESTATION_ALLOWED_SIGNATURE_ALGORITHMS)
+            val claimsSet = jwt.ensureValidClaimsSet<KeyAttestationJWTClaims>()
+            return KeyAttestationJWT(jwt.serialize(), jwt.header, claimsSet)
+        }
     }
+}
 
-    companion object {
+fun KeyAttestationJWT.serialize(): String = jwt
```
vafeini: Is this extension fun needed? Why not a simple get() on jwt property?

src/main/kotlin/eu/europa/ec/eudi/openid4vci/Issuance.kt
```diff
@@ -146,15 +146,18 @@ sealed interface ProofsSpecification {
...
+        }
     }
 
     data class AttestationProof(
         val attestationProvider: suspend (Nonce?) -> KeyAttestationJWT,
     ) : ProofsSpecification
 }
 
+internal operator fun ProofsSpecification.JwtProofs.WithKeyAttestation.component2(): Int = keyIndex
```
vafeini: No need for this and the keyIndex. It is a given that the signing key must be the first one in attested_keys and it is mandatory that kid value must be 0. It is not an arbitrary index that might change. 

If there is a check that must be performed it should be that the signer
passed in `KeyAttestationJwtProofSigner` is using the first key from the list of attested_keys

src/main/kotlin/eu/europa/ec/eudi/openid4vci/KeyAttestationJWT.kt
dzarras: Converted to a value class instead that enforces the invariants that were part of the KeyAttestationJWTClaims init block.

src/main/kotlin/eu/europa/ec/eudi/openid4vci/KeyAttestationJWT.kt
dzarras: Done, please check again.

src/main/kotlin/eu/europa/ec/eudi/openid4vci/StatusClaim.kt
dzarras: By spec:

> uri: REQUIRED. The uri (URI) claim MUST specify a String value that identifies the Status List Token containing the status information for the Referenced Token. The value of uri MUST be a URI conforming to [[RFC3986](https://www.ietf.org/archive/id/draft-ietf-oauth-status-list-20.html#RFC3986)].

Updated the type to `URI` instead and introduced a new `Serializer`

src/main/kotlin/eu/europa/ec/eudi/openid4vci/ClientAttestation.kt
dzarras: Removed.

src/main/kotlin/eu/europa/ec/eudi/openid4vci/ClientAttestation.kt
dzarras: Done

src/main/kotlin/eu/europa/ec/eudi/openid4vci/Issuance.kt
dzarras: Removed.

src/main/kotlin/eu/europa/ec/eudi/openid4vci/internal/RequestIssuanceImpl.kt
dzarras: Parsing of the KeyAttestationJWT would have failed if it is not signed using ECDSA.
Nonetheless updated the code to require EC Keys.

src/main/kotlin/eu/europa/ec/eudi/openid4vci/StatusClaim.kt
dzarras: Moved to `Types.kt`

src/main/kotlin/eu/europa/ec/eudi/openid4vci/KeyAttestationJWT.kt
dzarras: Removed.

src/main/kotlin/eu/europa/ec/eudi/openid4vci/Issuance.kt
dzarras: The property has been removed. A constant is used in its place wherever needed.

vafeini (approved):

babisRoutis (approved):

PR #535, opened 2026-05-13
Add support for parsing and providing `preferred_key_storage_status_period`

### LINKED ISSUE #534: Parse and provide `preferred_key_storage_status_period`
TS3 introduces a new claim in `key_attestations_required`:

> `preferred_key_storage_status_period`: **OPTIONAL**. An integer specifying a PID or Attestation Provider's preference for the remaining status maintenance period (`key_storage_status.exp` minus current time) of the KA it receives during issuance, in seconds. See [Section 2.4.2](https://github.com/eu-digital-identity-wallet/eudi-doc-standards-and-technical-specifications/blob/main/docs/technical-specifications/ts3-wallet-unit-attestation.md/#242-wallet-provider-responsibilities) for how the Wallet Unit and the Wallet Provider must handle this parameter.

We must update the library to parse this new claim and provide it to Wallets during `KeyAttestationJWT` provisioning.

dzarras:
Closes #534

### DIFF
--- a/src/main/kotlin/eu/europa/ec/eudi/openid4vci/CredentialConfiguration.kt
+++ b/src/main/kotlin/eu/europa/ec/eudi/openid4vci/CredentialConfiguration.kt
@@ -101,6 +101,7 @@ sealed interface KeyAttestationRequirement {
     data class Required(
         val keyStorageConstraints: List<String>?,
         val userAuthenticationConstraints: List<String>?,
+        val preferredKeyStorageStatusPeriod: PositiveDuration?,
     ) : KeyAttestationRequirement {
         init {
 
@@ -122,7 +123,7 @@ sealed interface KeyAttestationRequirement {
         companion object
     }
     companion object {
-        val RequiredNoConstraints: Required = Required(null, null)
+        val RequiredNoConstraints: Required = Required(null, null, null)
     }
 }
 
--- a/src/main/kotlin/eu/europa/ec/eudi/openid4vci/CredentialConfiguration.kt
+++ b/src/main/kotlin/eu/europa/ec/eudi/openid4vci/CredentialConfiguration.kt
@@ -122,9 +122,8 @@ sealed interface KeyAttestationRequirement {
 
         companion object
     }
-    companion object {
-        val RequiredNoConstraints: Required = Required(null, null, null)
-    }
+
+    companion object
 }
 
 fun ProofTypeMeta.type(): ProofType? = when (this) {
--- a/src/main/kotlin/eu/europa/ec/eudi/openid4vci/CredentialConfiguration.kt
+++ b/src/main/kotlin/eu/europa/ec/eudi/openid4vci/CredentialConfiguration.kt
@@ -99,8 +99,8 @@ sealed interface KeyAttestationRequirement {
     }
 
     data class Required(
-        val keyStorageConstraints: List<String>?,
-        val userAuthenticationConstraints: List<String>?,
+        val keyStorageConstraints: List<AttackPotentialResistance>?,
+        val userAuthenticationConstraints: List<AttackPotentialResistance>?,
         val preferredKeyStorageStatusPeriod: PositiveDuration?,
     ) : KeyAttestationRequirement {
         init {
--- a/src/main/kotlin/eu/europa/ec/eudi/openid4vci/internal/RequestIssuanceImpl.kt
+++ b/src/main/kotlin/eu/europa/ec/eudi/openid4vci/internal/RequestIssuanceImpl.kt
@@ -108,18 +108,20 @@ internal class RequestIssuanceImpl(
         credentialConfigId: CredentialConfigurationIdentifier,
         grant: Grant,
     ): Pair<List<Proof>, Nonce?> {
-        proofsSpecification.ensureCompatibleWith(credentialConfigId)
+        val credentialConfiguration = credentialSupportedById(credentialConfigId)
+        val proofRequirement = proofsSpecification.ensureCompatibleWith(credentialConfiguration)
 
         return when (proofsSpecification) {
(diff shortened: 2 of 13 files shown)


src/main/kotlin/eu/europa/ec/eudi/openid4vci/internal/RequestIssuanceImpl.kt
```diff
@@ -128,24 +130,26 @@ internal class RequestIssuanceImpl(
...
                     selectedReusePolicy,
-                    credentialConfigId,
                     grant,
                     cNonceAndDPoPNonce?.cnonce,
                 )
                 proofs to cNonceAndDPoPNonce?.dpopNonce
             }
 
             is ProofsSpecification.AttestationProof -> {
+                check(proofRequirement is ProofTypeMeta.Attestation)
```
vafeini: Isn't this check already performed in proofsSpecification.ensureCompatibleWith(credentialConfiguration)?

src/main/kotlin/eu/europa/ec/eudi/openid4vci/internal/RequestIssuanceImpl.kt
```diff
@@ -128,24 +130,26 @@ internal class RequestIssuanceImpl(
             }
 
             is ProofsSpecification.JwtProofs.NoKeyAttestation -> {
+                check(proofRequirement is ProofTypeMeta.Jwt)
```
vafeini: Isn't this check already performed in proofsSpecification.ensureCompatibleWith(credentialConfiguration)?

src/main/kotlin/eu/europa/ec/eudi/openid4vci/internal/RequestIssuanceImpl.kt
```diff
@@ -108,18 +108,20 @@ internal class RequestIssuanceImpl(
...
     ): Pair<List<Proof>, Nonce?> {
-        proofsSpecification.ensureCompatibleWith(credentialConfigId)
+        val credentialConfiguration = credentialSupportedById(credentialConfigId)
+        val proofRequirement = proofsSpecification.ensureCompatibleWith(credentialConfiguration)
 
         return when (proofsSpecification) {
             is ProofsSpecification.NoProofs -> emptyList<Proof>() to null
 
             is ProofsSpecification.JwtProofs.WithKeyAttestation -> {
+                check(proofRequirement is ProofTypeMeta.Jwt)
```
vafeini: Isn't this check already performed in proofsSpecification.ensureCompatibleWith(credentialConfiguration)?

src/main/kotlin/eu/europa/ec/eudi/openid4vci/internal/RequestIssuanceImpl.kt
dzarras: true. they were used to trigger a smart cast by the compiler. replaced with explicit cast.

src/main/kotlin/eu/europa/ec/eudi/openid4vci/internal/RequestIssuanceImpl.kt
dzarras: true. they were used to trigger a smart cast by the compiler. replaced with explicit cast.

src/main/kotlin/eu/europa/ec/eudi/openid4vci/internal/RequestIssuanceImpl.kt
dzarras: true. they were used to trigger a smart cast by the compiler. replaced with explicit cast.

src/main/kotlin/eu/europa/ec/eudi/openid4vci/CredentialConfiguration.kt
```diff
@@ -101,6 +101,7 @@ sealed interface KeyAttestationRequirement {
     data class Required(
         val keyStorageConstraints: List<String>?,
```
babisRoutis: I think there was a type for the contents of the list.
No?

src/main/kotlin/eu/europa/ec/eudi/openid4vci/CredentialConfiguration.kt
```diff
@@ -122,7 +123,7 @@ sealed interface KeyAttestationRequirement {
         companion object
     }
     companion object {
-        val RequiredNoConstraints: Required = Required(null, null)
+        val RequiredNoConstraints: Required = Required(null, null, null)
```
babisRoutis: Is this useful? Is there a real use for this?

How is this different from `NotRequired`

src/main/kotlin/eu/europa/ec/eudi/openid4vci/CredentialConfiguration.kt
dzarras: Replaced with `AttackPotentialResistance` from `KeyAttestationJWT.kt`.

src/main/kotlin/eu/europa/ec/eudi/openid4vci/CredentialConfiguration.kt
dzarras: `NotRequired` means no Key Attestation is required.

`RequiredNoConstraints` means Key Attestation is required, but the Credential Issuer does not enforce constraints for either `user_authentication`, or `key_storage`, and has no `preferred_key_storage_status_period`.

I'll move `RequiredNoConstraints` to the test target, since it is were we use it.

dzarras: It was being used in a single test case. I removed it and kept an empty companion object so that users are still able to define extensions.

PR #560, opened 2026-07-15
Fix Credential Reuse Policy parsing

dzarras:
This PR updates the parsing of Credential Reuse Policies and moves the following checks at the `CredentialReusePolicy.EUDI` level rather than at the `EudiReusePolicy` level:

1. only unique details are provided
2. details contains either `once_only` or `limited-time`

With the above changes, the library is now able to successfully parse Credential Reuse Policies like below:
```json
        "credential_reuse_policy": {
          "id": "arf_annex_ii",
          "options": [
            {
              "details": ["limited_time"],
              "batch_size": 40,
              "reissue_trigger_unused": 10,
              "reissue_trigger_lifetime_left": 655433
            },
            {
              "details": ["rotating-batch"],
              "batch_size": 40,
              "reissue_trigger_unused": 10,
              "reissue_trigger_lifetime_left": 655433
            },
            {
              "details": ["per-relying-party"],
              "batch_size": 40,
              "reissue_trigger_unused": 10,
              "reissue_trigger_lifetime_left": 655433
            }
          ]
        }
```

Additionally, this PR adds extra test cases to verify that:
1. When duplicates are encountered, parsing fails
2. When both `once_only`, and `limited-time` is present, parsing fails
3. When none `once_only`, nor `limited-time` is present, parsing fails

Tests cover both compact and expanded forms.

### DIFF
--- a/src/main/kotlin/eu/europa/ec/eudi/openid4vci/CredentialReusePolicy.kt
+++ b/src/main/kotlin/eu/europa/ec/eudi/openid4vci/CredentialReusePolicy.kt
@@ -124,15 +124,9 @@ sealed interface EudiReusePolicy {
             reissueTriggerUnused: Int? = null,
             reissueTriggerLifetimeLeft: Duration? = null,
         ): List<EudiReusePolicy> {
-            val normalizedDetails = details.distinct()
+            require(details.isNotEmpty()) { "details must not be empty" }
 
-            require(normalizedDetails.isNotEmpty()) { "details must not be empty" }
-            require(normalizedDetails.size == details.size) {
-                "details must not contain duplicate values"
-            }
-            validateBaseMethodCombination(normalizedDetails)
-
-            return normalizedDetails.map { detail ->
+            return details.map { detail ->
                 when (detail) {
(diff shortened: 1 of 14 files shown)


src/main/kotlin/eu/europa/ec/eudi/openid4vci/CredentialReusePolicy.kt
```diff
@@ -124,15 +124,9 @@ sealed interface EudiReusePolicy {
             reissueTriggerUnused: Int? = null,
             reissueTriggerLifetimeLeft: Duration? = null,
         ): List<EudiReusePolicy> {
-            val normalizedDetails = details.distinct()
+            require(details.isNotEmpty()) { "details must not be empty" }
 
-            require(normalizedDetails.isNotEmpty()) { "details must not be empty" }
```
vafeini, line 129: We still need this check here when parsing each array element of policies array

dzarras: Check has been re-introduced.

vafeini (approved):

PR #564, opened 2026-07-20
Treat key_attestations_required as optional when parsing proof types

15characterlimi:
https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html says: *`key_attestations_required`: OPTIONAL*.

Further, absent is different from empty (both are allowed): *If the Credential Issuer does not require a key attestation, this parameter MUST NOT be present in the metadata. If both key_storage and user_authentication parameters are absent, the key_attestations_required parameter may be empty, indicating a key attestation is needed without additional constraints.*

The metadata parser in this library previously (wrongly) asserted that key_attestations_required is present.

This PR fixes this.

### DIFF
--- a/src/main/kotlin/eu/europa/ec/eudi/openid4vci/CredentialConfiguration.kt
+++ b/src/main/kotlin/eu/europa/ec/eudi/openid4vci/CredentialConfiguration.kt
@@ -64,7 +64,13 @@ enum class ProofType : Serializable {
 sealed interface ProofTypeMeta : Serializable {
     data class Jwt(
         val algorithms: List<JWSAlgorithm>,
-        val keyAttestationRequirement: KeyAttestationRequirement,
+        /**
+         * The issuer's key attestation requirement, or `null` when the metadata omits
+         * `key_attestations_required`. That parameter is OPTIONAL in OpenID4VCI; its absence means
+         * the issuer does not require a key attestation, which is distinct from an empty object
+         * (a key attestation is required, but without additional constraints).
+         */
+        val keyAttestationRequirement: KeyAttestationRequirement?,
     ) : ProofTypeMeta {
         init {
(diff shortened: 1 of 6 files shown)


dzarras:
Dear @15characterlimi ,

Thank you for reaching out and for your contribution.


In the README it's mentioned that:

> This is a Kotlin library, targeting JVM, that supports the [OpenId4VCI 1.0](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html) protocol, focusing on compliance with:
>
> * [ARF's Technical Specification 3](https://github.com/eu-digital-identity-wallet/eudi-doc-standards-and-technical-specifications/blob/main/docs/technical-specifications/ts3-wallet-unit-attestation.md)
> * [ETSI TS 119 472-3 V1.1.1 (2026-03)](https://www.etsi.org/deliver/etsi_ts/119400_119499/11947203/01.01.01_60/ts_11947203v010101p.pdf)


Per [ETSI TS 119 472-3 V1.1.1 (2026-03)](https://www.etsi.org/deliver/etsi_ts/119400_119499/11947203/01.01.01_60/ts_11947203v010101p.pdf) the supported Proof Types are:

1. JWT Proofs with Key Attestation
2. Attestation Proofs

Plain JWT Proofs, i.e. JWT Proofs without Key Attestation, are not supported.

It's a conscious decision to have `CredentialIssuerMetadataJsonParser` fail when it encounters a Credential Issuer that advertises plain JWT Proofs.


The release notes for [v0.12.0](https://github.com/eu-digital-identity-wallet/eudi-lib-jvm-openid4vci-kt/releases/tag/v0.12.0) contain an extensive list of Breaking Changes that have been introduced to the library to achieve compliance with the aforementioned specifications.

As such, unfortunately we cannot accept this PR.

Please let us know if you need further assistance.

Kind regards.

PR #565, opened 2026-07-22
Support WRP Registration Certificate 

### LINKED ISSUE #563: Support WRP Registration Certificate (WRPRC)
ETSI TS 119 475 V1.2.1 in section 4.5 defines the relationship of WRPRC and WRPAC as

> WRPRC and WRPAC serve distinct but complementary purposes within the EUDIW ecosystem. While the WRPAC
ensures WRP authentication, the WRPRC conveys the WRP's declared use cases and data access policies to both the
EUDIW and the end user. Together, they form a dual-layer trust framework, where the WRPAC guarantees the WRP's
identity and authentication, and the WRPRC ensures entitlements, transparency and data minimization.

Currently the access certificate trust validation is performed upon issuer metadata resolution (`DefaultCredentialIssuerMetadataResolver` component) and is controlled by the `IssuerMetadataPolicy` configuration class. 

What we are missing is the verification that the content of an issuance transaction is aligned with the registered WRPRC of the involved issuer. Currently library facilitates 2 cases of issuance transactions: (1) via a credential offer and (2) wallet initiated issuance where the content of the issuance transaction is decided by the wallet. Both of these converge into the resolution of a credential offer via `CredentialOffer.resolve()` or `CredentialOffer.walletInitiated()`.  

Having a resolved `CredentialOffer` library guaranties that:
- Issuer at hand has been authenticated (WRPAC trust validated)
- Issuer's metadata have been resolved and validated 
- The content of the issuance transaction is known and verified to be supported by issuer

Thus WRPRC policy validation must be performed upon a resolved `CredentialOffer`. Currently credential offer resolution is performed upon the creation of the `Issuer`  component via construction methods `Issuer.makeWalletInitiated()` or `Issuer.make()` that if successfull returns an instance of an `Issuer` that is capable of requesting the content of the issuance transaction. 
WRPRC validation should be injected in this process so that we can claim that the successfull creation of an `Issuer` also guaranties that WRPRC  is met (or met with warnings).

For the above to be implemented the following need to be done:
- [x] Parse `issuer_info` attribute of issuer metadata as defined in ETSI TS 119 472-3 v1.1.1 and CIR 2024/2082 Annex I and reflect it to `CredentialIssuerMetadata` domain class.
- [x] Introduce `RegistrationCertificatePolicy` interface that implements the policy evaluation. It will be a functional interface with a suspendable function that takes as input
      1. The WRPAC
      2. The WRPRC passed as a string
      3. A list of `CredentialConfiguration`s of the credentials that are in the context of the credential offer
- [x] The application of the policy should return an `Authorization` that will be a sealed hierarchy of 2 members:
       - `Authorization.Granted`: Denotes successfull evaluation of the policy with possible warnings expressed as a list of `PolicyViolation` value objects.
       - `Authorization.NotGranted`: Denotes that policy was violated and processing should stop. 
- [x] Add `RegistrationCertificatePolicy` as a nullable attribute of `OpenId4VCIConfig`. If not set, no policy evaluation should be performed
- [x] Intoduce a `RegistrationCertificatePolicyEvaluator` component that is initialized with a `RegistrationCertificatePolicy` and given a `CredentialOffer` does the folowing:
       - Extracts the WRPRC from `IssuerMetadata.issuerInfo`. Fails with error if not present.
       - No structural check is performed on the extracted WRPRC.
       - Apply WRPRC policy validation using `RegistrationCertificatePolicy`.
- [x] Update `Issuer.make()` to:
       1. Apply the WRPRC policy evaluation if a `RegistrationCertificatePolicy` is provided in configuration
       2. Return the list of possible policy violation warnings along with the instantiated `Issuer` as a `IssuerNegotiationResult`. The return value should be a pair of an `Issuer` and a `List<RegistrationCertificatePolicy.PolicyViolation>`

vafeini:
- Introduced `RegistrationCertificatePolicyEvaluator` for WRPRC policy evaluation
- `Issuer` factory methods return an `IssuerResolutionResult` sealed hierarchy that when successful might include warnings regarding the WRPRC policy violations
- Removed `IssuerTrust.ByPublicKey`, `IssuerTrust` is not a data ckass constructed with a `CertificateChainTrust`

Closes #563


src/main/kotlin/eu/europa/ec/eudi/openid4vci/Types.kt
```diff
@@ -370,17 +370,7 @@ fun interface CertificateChainTrust {
...
-    @Deprecated("Use IssuerTrust.ByCertificateChain instead")
-    data class ByPublicKey(val jwk: JWK) : IssuerTrust {
-        init {
-            require(!jwk.isPrivate) { "Only public JWKs are supported" }
-        }
-    }
-
-    data class ByCertificateChain(val certificateChainTrust: CertificateChainTrust) : IssuerTrust
-}
+data class IssuerTrust(val certificateChainTrust: CertificateChainTrust)
```
babisRoutis, line 374: Do we really need this class?

I would say
a) Deprecate it, with a hint to use CertificateChainTrust
b) Add methods (deprecated) to bridge it to CertificateChainTrust on the calling side


There are some data classes that use this, that can use `CertificateChainTrust` with the addition of a 2nd constructor (deprecated) that uses IssuerTrust

src/main/kotlin/eu/europa/ec/eudi/openid4vci/Specs.kt
```diff
@@ -189,3 +193,7 @@ object TokenStatusListSpec {
     const val INDEX: String = "idx"
     const val URI: String = "uri"
 }
+
+object ETSI119475 {
```
babisRoutis: This is used only in tests I think

We don't want to "introspect" the JWT/CWT, as protocol library.

Consider moving it to tests

src/main/kotlin/eu/europa/ec/eudi/openid4vci/Issuer.kt
```diff
@@ -26,6 +26,42 @@ import kotlinx.coroutines.coroutineScope
 import java.net.URI
 import java.time.Instant
 
+/**
+ * Represents the result of an issuer resolution process which can either be successful or a failure.
+ */
+sealed interface IssuerResolutionResult {
```
babisRoutis: I think that "Resolution", as a term, has been overused 


What about `IssuerNegotiationResult`?
Mental model: You can get a "facade" of the issuer after "negotiating" on policies

src/main/kotlin/eu/europa/ec/eudi/openid4vci/Issuer.kt
```diff
@@ -26,6 +26,42 @@ import kotlinx.coroutines.coroutineScope
...
+
+    /**
+     * Represents a failed outcome of the issuer resolution process.
+     *
+     * This class encapsulates details about the failure that occurred during the issuer resolution process.
+     * The failure is represented by an associated exception, providing additional context about the error.
+     *
+     * @property exception The exception encountered during the issuer resolution process.
+     */
+    data class Failure(
```
babisRoutis: It is not clear what this non-recoverable state is.

I guess it represents a condition on  the "negotiation" that was not met (and it is being reported as a Throwable)

Perhaps, there should be a dedicated state to express the fact that the Issuer is not compliant to the policy of the Wallet Provider.


The 1st is technical aspect. The 2nd (if introduced) is a policy violation


Of course counter argument could be that WRPRC policy violations have a dedicated exception

`AuthorizationPolicyValidationError.AuthorizationPolicyNotMet`




On the other hand, should we keep the exception approach:

The previous resulting value was Result<Issuer>. Perhaps we can just to use the following

```kotlin


typealias IssuerResolutionResult = Result<Pair<Issuer, List<RegistrationCertificatePolicy.PolicyViolation>>>

```

src/main/kotlin/eu/europa/ec/eudi/openid4vci/Issuer.kt
```diff
@@ -423,3 +481,28 @@ fun ProvisionClientAttestation.ensureValid(now: Instant, provisioned: ProvisionC
...
+    class MissingIssuerInfo :
+        AuthorizationPolicyValidationError(IllegalArgumentException("Missing issuer info"))
+
+    class MissingRegistrationCertificate :
+        AuthorizationPolicyValidationError(IllegalArgumentException("Missing issuer registration certificate"))
+
+    class MultipleRegistrationCertificates :
+        AuthorizationPolicyValidationError(IllegalArgumentException("Multiple Registration Certificates provided while only one expected"))
+
+    class MissingAccessCertificate :
```
babisRoutis, line 453: Is this an Authorization or Authentication error?

The presense of the WRPAC is not governed by WRPRC option

src/main/kotlin/eu/europa/ec/eudi/openid4vci/Config.kt
```diff
@@ -122,8 +154,21 @@ data class OpenId4VCIConfig(
     val issuerMetadataPolicy: IssuerMetadataPolicy = IssuerMetadataPolicy.IgnoreSigned,
     val supportedCredentialReusePolicies: CredentialReusePolicies? = null,
     val proofs: ProofsConfig = ProofsConfig.Default,
+    val registrationCertificatePolicy: RegistrationCertificatePolicy? = null,
 ) {
 
+    init {
+        if (registrationCertificatePolicy != null) {
+            ensure(issuerMetadataPolicy is IssuerMetadataPolicy.RequireSigned) {
+                IllegalStateException(
```
babisRoutis: Minor:

Typically, this should be `IllegalArgumentException`. A data class has only a single state.

src/main/kotlin/eu/europa/ec/eudi/openid4vci/CredentialIssuerMetadataResolver.kt
```diff
@@ -161,7 +216,9 @@ data class CredentialIssuerMetadata(
     val credentialConfigurationsSupported: Map<CredentialConfigurationIdentifier, CredentialConfiguration>,
     val display: List<Display> = emptyList(),
     val preferredClientStatusPeriod: PositiveDuration? = null,
-) : Serializable {
+    val issuerInfo: IssuerInfo? = null,
+    val accessCertificate: X509Certificate? = null,
```
babisRoutis: Minor

I have some 2nd thoughts on the name `accessCertificate`. Actually, for the library is just the certificate discovered in the `x5c` of the singed credential issuer metadata, with no further strings attached.

The current may give the wrong impression that library checks/enforces the certificate to actually be an WRPAC.

src/test/kotlin/eu/europa/ec/eudi/openid4vci/examples/PidDevIssuer.kt
```diff
@@ -48,6 +48,7 @@ internal object PidDevIssuer :
         parUsage = ParUsage.Required(),
         supportedCredentialReusePolicies = CredentialReusePolicies.Supported(setOf(EudiReusePolicyType.OnceOnly)),
         proofs = ProofsConfig.Default,
+        issuerMetadataPolicy = IssuerMetadataPolicy.RequireSigned(IssuerTrust({ true })),
```
babisRoutis: Add a comment like DO NOT USE IN PRODUCTION

src/test/kotlin/eu/europa/ec/eudi/openid4vci/DefaultCredentialIssuerMetadataResolverTest.kt
```diff
@@ -549,7 +534,7 @@ internal class DefaultCredentialIssuerMetadataResolverTest {
                 listOf("application/jwt"),
             ),
         )
-        val policy = IssuerMetadataPolicy.RequireSigned(IssuerTrust.ByCertificateChain { true })
+        val policy = IssuerMetadataPolicy.RequireSigned(IssuerTrust({ true }))
```
babisRoutis: Please replace the occureancies of IssuerTrust {true } with  a single val TrustAll = IssuerTrust { true} with a comment that this is only for tests 


When test are done you can even use the actual certificate of WrpacProvider test utility

src/test/kotlin/eu/europa/ec/eudi/openid4vci/IssuerTest.kt
```diff
@@ -15,14 +15,85 @@
...
-import kotlin.test.assertNotNull
-import kotlin.test.assertNull
+import java.net.URI
+import kotlin.test.*
 
 class IssuerTest {
 
+    @Test
+    fun `when wrprc policy provided then IssuerMetadataPolicy must be RequireSigned`() = runTest {
+        assertFailsWith<IllegalStateException> {
```
babisRoutis: Should be changed to IllegalArgumentException

src/main/kotlin/eu/europa/ec/eudi/openid4vci/Issuer.kt
vafeini: The `typealias` approach was my initial implementation. Though during issuer negotiation we throw several business errors (ex. offer resolution failures due to failing checks, incompatibilities with auth server) and thought it would make sense to represent it with a specific resolution state.

On the other hand the `Result`'s error channel is used widely in our implementation to communicate errors. Maybe it makes sense to keep it consistent also in the issuer construction case. 

Will follow the `typealias IssuerResolutionResult = Result<Pair<>` approach

src/main/kotlin/eu/europa/ec/eudi/openid4vci/CredentialIssuerMetadataResolver.kt
vafeini: Shall we call it `metadataSigningCertificate` ?

src/main/kotlin/eu/europa/ec/eudi/openid4vci/Issuer.kt
vafeini: It is a check performed during authorization (WRPRC policy evaluation) that issuer should be authenticated (has a WRPAC). That is why it is reported as and authorization error. 

Shall we create another hierarchy of Authentication errors with `MissingAccessCertificate` as a member?

src/main/kotlin/eu/europa/ec/eudi/openid4vci/Specs.kt
vafeini: Done

src/main/kotlin/eu/europa/ec/eudi/openid4vci/Issuer.kt
vafeini: Done

src/main/kotlin/eu/europa/ec/eudi/openid4vci/Issuer.kt
vafeini: done

src/main/kotlin/eu/europa/ec/eudi/openid4vci/Config.kt
vafeini: done

src/main/kotlin/eu/europa/ec/eudi/openid4vci/CredentialIssuerMetadataResolver.kt
vafeini: Rename to `metadataSigningCertificate`

src/test/kotlin/eu/europa/ec/eudi/openid4vci/IssuerTest.kt
vafeini: Done

src/test/kotlin/eu/europa/ec/eudi/openid4vci/examples/PidDevIssuer.kt
vafeini: done

src/test/kotlin/eu/europa/ec/eudi/openid4vci/DefaultCredentialIssuerMetadataResolverTest.kt
vafeini: done

src/main/kotlin/eu/europa/ec/eudi/openid4vci/Types.kt
vafeini: Done

PR #569, opened 2026-07-31
Fix incorrect handling of issuer who only supports one of the two ProofTypes (JWT or Attestation), but not both.

15characterlimi:
Fix incorrect handling of issuer who only supports one of the two ProofTypes (JWT or Attestation), but not both.
    
Commit 91de4f9b264b44202853f916635e8a59f1cca416 had (rightly) relaxed an assertion that previously required that issuers support both, but this code here was still unconditionally casting to non-nullable ProofTypeMeta.Jwt and ProofTypeMeta.Attestation, either of which might be null.

This PR consists of two commits:
 1. A test demonstrating the bug
 2. The fix.

To demonstrate the fix, the following fails after the first commit but passes after the second commit:
    
 `./gradlew test --tests "eu.europa.ec.eudi.openid4vci.internal.ProofsConfigCompatibilityTest"`
 
 Note: for any further discussions or pushback / needed changes, please contact @spacifici or @simplekjl rather than myself.


dzarras:
Dear @15characterlimi,  @spacifici, @simplekjl,

What you are describing is indeed a bug.
Thank you a lot for your contribution. 
We took the liberty of further refining the code a bit.

We will review it and provide a bugfix release promptly.

Kind regards.

vafeini (approved):

PR #576, opened 2026-08-24
Require Authorization Code Grant, and Pre-Authorized Code Grant use the same Authorization Server when Grants is Both

### LINKED ISSUE #575: Grants.Both with different authorization_servers can send pre-authorized code to the wrong token endpoint
When a credential offer has both an `authorization_code` and a pre-`authorized_code` grant naming different `authorization_servers`, the resolver commits to a single server (preferring the authorization-code grant). Both flows are then bound to that one server's token endpoint. If the pre-authorized flow runs, the pre-authorized_code, transaction code, and related authentication data are sent to the authorization-code grant's server instead of the pre-authorized grant's server.

**Root cause**
`DefaultCredentialOfferRequestResolver.kt:
`
```
private fun Grants.authServer(): HttpsUrl? = when (this) {
    is Grants.AuthorizationCode -> authorizationServer
    is Grants.PreAuthorizedCode -> authorizationServer
    is Grants.Both -> authorizationCode.authorizationServer ?: preAuthorizedCode.authorizationServer
}
```
`val authorizationServer = grants?.authServer() ?: credentialIssuerMetadata.authorizationServers[0]`
The Both case collapses to one server, and the resolved `authorizationServerMetadata` is stored as a single field on `CredentialOffer`. Issuer.make builds its authorization and token clients against that one server, so the flow the caller later picks may not match the grant whose server was selected.

Example
```
"grants": {
  "authorization_code":  { "authorization_server": "https://as-A.example" },
  "pre-authorized_code": { "pre-authorized_code": "SECRET",
                           "tx_code": { "length": 6 },
                           "authorization_server": "https://as-B.example" }
}
```
Resolver picks as-A; running the pre-authorized flow sends SECRET + tx code to as-A, not as-B.

Recommendation
Keep authorization-server metadata per grant and select the token endpoint only after the flow is chosen — or reject offers whose two grants name different authorization servers.

(The equivalent iOS library eudi-lib-ios-openid4vci-swift has the same flaw.)

dzarras:
This PR introduces a new constraint in `Grants.Both` and enforces both `AuthorizationCode`, and `PreAuthorizedCode` use the same Authorization Server.

Closes #575


src/main/kotlin/eu/europa/ec/eudi/openid4vci/CredentialOffer.kt
```diff
@@ -166,7 +166,15 @@ sealed interface Grants : Serializable {
     data class Both(
         val authorizationCode: AuthorizationCode,
         val preAuthorizedCode: PreAuthorizedCode,
-    ) : Grants
+    ) : Grants {
+        init {
+            if (null != authorizationCode.authorizationServer && null != preAuthorizedCode.authorizationServer) {
```
babisRoutis, line 174: @dzarras Please a comment to the source code, elaborating why this conditions is being required

babisRoutis:
@dzarras 

Can you please consider if would it make sense to introduce a new configuration option, something like supported grants in the form of a sealed hierarchy with the following members:

- AuthorizationCodeGrant
- PreAuthorizedCode
- Both

Having as default the value `Both`

In addition we could mark the configuration option " val authFlowRedirectionURI: URI" as optional, only required if configuration is either `AuthorizationCodeGrant` or `Both`

This would allow the caller to explicitly provide a "hint" to the credential offer resolution function. 
Of course, this could be a part of another PR. In any case the  restriction introduced by this PR will remain

src/main/kotlin/eu/europa/ec/eudi/openid4vci/CredentialOffer.kt
babisRoutis: Probably we should document this library restriction to the README.md

dzarras: Done

dzarras:
> @dzarras
> 
> Can you please consider if would it make sense to introduce a new configuration option, something like supported grants in the form of a sealed hierarchy with the following members:
> 
> * AuthorizationCodeGrant
> * PreAuthorizedCode
> * Both
> 
> Having as default the value `Both`
> 
> In addition we could mark the configuration option " val authFlowRedirectionURI: URI" as optional, only required if configuration is either `AuthorizationCodeGrant` or `Both`
> 
> This would allow the caller to explicitly provide a "hint" to the credential offer resolution function. Of course, this could be a part of another PR. In any case the restriction introduced by this PR will remain

I think it makes sense. It would allow us to stop the resolution and raise an error.

Let's tackle this in a separate PR.

README.md
```diff
@@ -156,6 +156,13 @@ This resolution includes the following
 - Fetch and validate the OAuth 2.0 Authorization Server metadata used by the Credential Issuer
 - Ensure that credential offer is aligned to both the above metadata & wallet issuance configuration
 
+> [!IMPORTANT]
+>
+> The library currently supports resolving Credential Offers that contain Grants that use the use Authorization Server.
```
babisRoutis: There is a minor type "that use the use Authorization Server"

babisRoutis (dismissed):
Just a typo

PR #578, opened 2026-08-24
Introduce configuration option for supported Grants

### LINKED ISSUE #577: Introduce a new configuration option to allow Wallets configure the supported Grant types
A new configuration option should be introduced to allow Wallet to configure the supported Grant Types.

The new option can be an enumeration or sealed hierarchy with the following members:

1. AuthorizationCode
2. PreAuthorizedCode
3. Both

Default value can be `Both`. When `AuthorizationCode` or `Both` is used, `OpenId4VCIConfig` must contain `authFlowRedirectionURI`, otherwise it can be `null`.

The new configuration option must be used by `CredentialOfferRequestResolver` and raise an error when the resolved `CredentialOffer` doesn't contain a Wallet supported Grant.

dzarras:
Closes #577

### DIFF
--- a/src/main/kotlin/eu/europa/ec/eudi/openid4vci/internal/CredentialOfferRequestResolver.kt
+++ b/src/main/kotlin/eu/europa/ec/eudi/openid4vci/internal/CredentialOfferRequestResolver.kt
@@ -105,6 +105,7 @@ internal class CredentialOfferRequestResolver(
             }
 
             val grants = credentialOffer.grants?.toGrants(credentialIssuerMetadata)
+            grants?.ensureSupported(config.grants)
             val authorizationServer = grants?.authServer() ?: credentialIssuerMetadata.authorizationServers[0]
             val authorizationServerMetadata = fetchAuthServerMetaData(authorizationServer)
             val authorizationCodeGrant = grants?.authorizationCode()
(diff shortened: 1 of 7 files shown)


src/main/kotlin/eu/europa/ec/eudi/openid4vci/internal/CredentialOfferRequestResolver.kt
```diff
@@ -272,3 +273,25 @@ internal fun <A : Any, B : Any> DPoPUsage<A>.map(convert: (A) -> B): DPoPUsage<B
         is DPoPUsage.IfSupported -> DPoPUsage.IfSupported(convert(value))
         is DPoPUsage.Required -> DPoPUsage.Required(convert(value))
     }
+
+private fun Grants.ensureSupported(supported: SupportedGrants) {
```
babisRoutis: I would expect this function to return a "SupportedGrants" value. 

Perhaps, `SupportedGrants` can be renamed to simply `VciGrants`, in order to be used as a result type. 

The return value would represent the grants that are common between the wallet configuration and the credential offer.

dzarras: @babisRoutis Updated the logic of the function based on you suggestion. I haven't renamed `SupportedGrants` though. I feel `SupportedGrants` better conveys the intend compared to `VciGrants`.

What do you think?

vafeini (approved):

babisRoutis (approved):

PR #599, opened 2026-08-31
Fix 597

### LINKED ISSUE #597: OID4VCI Credential Issuer metadata discovery should preserve trailing slash in issuer path
### Summary

When resolving OID4VCI Credential Issuer metadata, the wallet appears to remove a trailing slash from the `credential_issuer` identifier path before constructing the well-known URL.

This causes interoperability issues with issuers and conformance tests that expect the `credential_issuer` path to be preserved exactly when building the Credential Issuer metadata URL.

### Context

While running the OID4VCI wallet conformance tests, I encountered a failure related to Credential Issuer metadata discovery.

The Credential Issuer Identifier used in the test ends with a trailing slash:
```
https://www.certification.openid.net/test/a/credimi-oid4vp-verifier-test-cf92628b-d782-4e6b-a116-5aad4897f9af/
````

The wallet appears to construct the metadata URL as:
```
https://www.certification.openid.net/.well-known/openid-credential-issuer/test/a/credimi-oid4vp-verifier-test-cf92628b-d782-4e6b-a116-5aad4897f9af
```

However, the expected URL is:
```
https://www.certification.openid.net/.well-known/openid-credential-issuer/test/a/credimi-oid4vp-verifier-test-cf92628b-d782-4e6b-a116-5aad4897f9af/
```

Image

### Relevant specification behavior

OID4VCI defines Credential Issuer metadata discovery by inserting:
```
/.well-known/openid-credential-issuer
```
between the host component and the path component of the Credential Issuer Identifier.

Therefore, for a Credential Issuer Identifier such as:
```
https://www.certification.openid.net/test/a/credimi-oid4vp-verifier-test-cf92628b-d782-4e6b-a116-5aad4897f9af/
```
the path component is:
```
/test/a/credimi-oid4vp-verifier-test-cf92628b-d782-4e6b-a116-5aad4897f9af/
```
and the resulting Credential Issuer metadata URL should preserve that path:
```
https://www.certification.openid.net/.well-known/openid-credential-issuer/test/a/credimi-oid4vp-verifier-test-cf92628b-d782-4e6b-a116-5aad4897f9af/
```

OID4VCI also requires the `credential_issuer` value returned in the metadata to be identical to the Credential Issuer Identifier used to create the metadata URL, using simple string comparison with no normalization. This suggests that the issuer identifier should not be normalized by removing the trailing slash during Credential Issuer metadata resolution.

### Difference from RFC 8414

[RFC 8414](https://datatracker.ietf.org/doc/html/rfc8414#section-3.1) defines OAuth Authorization Server metadata discovery and explicitly states that, when the issuer identifier contains a path component, any terminating `/` MUST be removed before inserting the well-known path.

That rule is appropriate for OAuth Authorization Server metadata

However, OID4VCI Credential Issuer metadata discovery does not repeat the RFC 8414 instruction to remove a terminating slash from the path.

As a result, RFC 8414 trailing-slash normalization should not be applied when resolving OID4VCI Credential Issuer metadata.

### Expected behavior

When resolving OID4VCI Credential Issuer metadata, the wallet should preserve the path component of the `credential_issuer` identifier exactly, including a trailing slash if present.

### Actual behavior

The wallet removes the trailing slash from the `credential_issuer` path.

### Impact

This affects interoperability with issuers and conformance tests where the Credential Issuer Identifier includes a trailing slash.

A wallet that normalizes the issuer path may fail to retrieve the correct Credential Issuer metadata document even though the issuer exposes it at the URL derived according to the OID4VCI Credential Issuer metadata discovery rules.

babisRoutis:
Closes #597

### DIFF
--- a/src/main/kotlin/eu/europa/ec/eudi/openid4vci/internal/DefaultCredentialIssuerMetadataResolver.kt
+++ b/src/main/kotlin/eu/europa/ec/eudi/openid4vci/internal/DefaultCredentialIssuerMetadataResolver.kt
@@ -197,6 +197,10 @@ internal fun CredentialIssuerId.wellKnown(): Url {
             OpenId4VCISpec.CREDENTIAL_ISSUER_WELL_KNOWN_PATH.trim('/').split("/"),
             encodeSlash = true,
         )
-        issuer.segments.forEach { appendPathSegments(listOf(it), encodeSlash = true) }
+        // Use `rawSegments` (not `segments`) so that the path component of the Credential Issuer
+        // Identifier is preserved exactly, including a trailing slash when present.
+        // OID4VCI Credential Issuer metadata discovery inserts the well-known path between the host
+        // and the path component without the RFC 8414 trailing-slash normalization.
+        issuer.rawSegments.forEach { appendPathSegments(listOf(it), encodeSlash = true) }
     }.build()
 }
--- a/src/test/kotlin/eu/europa/ec/eudi/openid4vci/DefaultCredentialIssuerMetadataResolverTest.kt
+++ b/src/test/kotlin/eu/europa/ec/eudi/openid4vci/DefaultCredentialIssuerMetadataResolverTest.kt
@@ -133,6 +133,33 @@ internal class DefaultCredentialIssuerMetadataResolverTest {
     }
 
     @Test
+    internal fun `resolution success preserves trailing slash in credential issuer path`() = runTest {
+        val credentialIssuerId = CredentialIssuerId(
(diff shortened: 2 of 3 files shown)


dzarras (changes requested):
~~@babisRoutis I think we have to update `DefaultAuthorizationServerMetadataResolver` similarly.~~

~~Reference: https://www.rfc-editor.org/info/rfc8414/#section-3~~

dzarras (approved):

dzarras:
@babisRoutis Ignore the above comment. PR approved.

PR #600, opened 2026-09-01
Fix 579

### LINKED ISSUE #579: The authorization response is accepted without the RFC 9207 `iss` check
HAIP mandates the usage of RFC9207: https://openid.net/specs/openid4vc-high-assurance-interoperability-profile-1_0.html#name-openid-for-verifiable-crede

1. The library must ensure that only Authorization Servers that support RFC 9207: OAuth 2.0 Authorization Server Issuer Identification are allowed, i.e. check that `authorization_response_iss_parameter_supported` is `true` in the advertised Authorization Server Metadata, 
2. The library must extract the `iss` query parameter from the Authorization Response, ensure it matches the `iss` value in the Authorization Server Metadata, store it, and then compare it to the value of `iss` in the Token Response.

babisRoutis:
Closes #579 

The PR implements the check described in RFC9207 with regards to the `iss` parameter, as follows:

A new parameter is introduced to the `OpenId4VCIConfig`,  named `authResponseIssChecking`. 
This new parameter can have one of the following values: Never, IfSupported, Required, and allows the caller to opt-in for the RFC9207 of the `iss` parameter, in case of Authorization Code Grant.

If this parameter is set to `Required`, the library would check that the Auhorization Server of the Credential Issuer advertises to its metadata a value for `authorization_response_iss_parameter_supported` equal to true.

In addition, the `AuthorizeIssuance` interface and especially the method 

```kotlin
suspend fun AuthorizationRequestPrepared.authorizeWithAuthorizationCode(
        authorizationCode: AuthorizationCode,
        serverState: String,
        authDetailsOption: AccessTokenOption,
        issuer: String? = null
): Result<AuthorizedRequest>
```
has been updated with the additional parameter `issuer` to allow the caller to pass the `iss` parameter for the redirection URI

Dedicated errors have been introduced : 
- `InvalidAuthorizationIssuer` when the `iss` is not the expected
- `MissingAuthorizationResponseIssuer` when AS misbehaves and doesn't return an `iss` (but advertises that it authorization_response_iss_parameter_supported)
- `AuthorizationResponseIssuerParamNotSupported` in case we have an authorization grant, and the wallet requires the `iss` parameter but the AS doesn't support it.

The above changes do affect the public API, in a non-breaking manner.

To enforce the rule, the caller has to
1) Explicitly set the `authResponseIssChecking` to `IfSupported` or `Required`
2) Extract from the redirection URI the `iss` parameter and call the `authorizeWithAuthorizationCode()` with the `iss` parameter.

Otherwise, the library should behave as previously

### DIFF
--- a/src/main/kotlin/eu/europa/ec/eudi/openid4vci/AuthorizeIssuance.kt
+++ b/src/main/kotlin/eu/europa/ec/eudi/openid4vci/AuthorizeIssuance.kt
@@ -115,15 +115,16 @@ interface AuthorizeIssuance {
      * [AuthorizedRequest] state.
      *
      * @param authorizationCode The authorization code returned from authorization server via front-channel
-     * @param serverState The state returned from authorization server via front-channel
+     * @param serverState The state returned from the authorization server via front-channel
      * @param authDetailsOption Defines if upon access token request extra authorization details will be set to fine grain the
      * scope of the access token.
-     * @return an issuance request in authorized state
+     * @return an issuance request in an authorized state
      */
     suspend fun AuthorizationRequestPrepared.authorizeWithAuthorizationCode(
         authorizationCode: AuthorizationCode,
         serverState: String,
         authDetailsOption: AccessTokenOption = AccessTokenOption.AsRequested,
+        issuer: String? = null,
     ): Result<AuthorizedRequest>
 
     /**
--- a/src/main/kotlin/eu/europa/ec/eudi/openid4vci/AuthorizeIssuance.kt
+++ b/src/main/kotlin/eu/europa/ec/eudi/openid4vci/AuthorizeIssuance.kt
@@ -116,8 +116,10 @@ interface AuthorizeIssuance {
      *
      * @param authorizationCode The authorization code returned from authorization server via front-channel
      * @param serverState The state returned from the authorization server via front-channel
-     * @param authDetailsOption Defines if upon access token request extra authorization details will be set to fine grain the
-     * scope of the access token.
+     * @param authDetailsOption Defines if upon access token request extra authorization details will be set to fine grain the scope of the access token.
+     * @param issuer The issuer of the access token, as provided to the redirection URI (the last step of authorization code grant) via parameter `iss`.
+     * It is optional and can be omited (defaulting to `null`) in case of Pre-Authorized Code Flow or if [OpenId4VCIConfig.authResponseIssChecking] is
+     * set to [AuthorizationResponseIssChecking.Never].
      * @return an issuance request in an authorized state
      */
     suspend fun AuthorizationRequestPrepared.authorizeWithAuthorizationCode(
--- a/src/main/kotlin/eu/europa/ec/eudi/openid4vci/internal/AuthorizeIssuanceImpl.kt
+++ b/src/main/kotlin/eu/europa/ec/eudi/openid4vci/internal/AuthorizeIssuanceImpl.kt
@@ -18,10 +18,14 @@ package eu.europa.ec.eudi.openid4vci.internal
 import eu.europa.ec.eudi.openid4vci.*
 import eu.europa.ec.eudi.openid4vci.AccessTokenOption.AsRequested
 import eu.europa.ec.eudi.openid4vci.AccessTokenOption.Limited
+import eu.europa.ec.eudi.openid4vci.CredentialIssuanceError.AuthorizationResponseIssuerParamNotSupported
+import eu.europa.ec.eudi.openid4vci.CredentialIssuanceError.InvalidAuthorizationIssuer
 import eu.europa.ec.eudi.openid4vci.CredentialIssuanceError.InvalidAuthorizationState
+import eu.europa.ec.eudi.openid4vci.CredentialIssuanceError.MissingAuthorizationResponseIssuer
 import eu.europa.ec.eudi.openid4vci.internal.http.AuthorizationEndpointClient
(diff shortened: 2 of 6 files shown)


src/main/kotlin/eu/europa/ec/eudi/openid4vci/AuthorizeIssuance.kt
```diff
@@ -115,15 +115,16 @@ interface AuthorizeIssuance {
...
      * @param authDetailsOption Defines if upon access token request extra authorization details will be set to fine grain the
      * scope of the access token.
-     * @return an issuance request in authorized state
+     * @return an issuance request in an authorized state
      */
     suspend fun AuthorizationRequestPrepared.authorizeWithAuthorizationCode(
         authorizationCode: AuthorizationCode,
         serverState: String,
         authDetailsOption: AccessTokenOption = AccessTokenOption.AsRequested,
+        issuer: String? = null,
```
dzarras, line 129: Please add KDocs for the new parameters.

src/main/kotlin/eu/europa/ec/eudi/openid4vci/internal/CredentialOfferRequestResolver.kt
```diff
@@ -106,14 +107,20 @@ internal class CredentialOfferRequestResolver(
...
+            val authorizationCodeGrant = grants?.authorizationCode()
+            if (authorizationCodeGrant != null) {
                 ensureNotNull(authorizationServerMetadata.authorizationEndpointURI) {
                     val error =
                         IllegalArgumentException(
                             "Credential Offer requires Authorization Code Grant, but the Authorization Server does not support it",
                         )
                     CredentialOfferRequestValidationError.InvalidGrants(error).toException()
                 }
+                if (config.authResponseIssChecking is AuthorizationResponseIssChecking.Required) {
```
dzarras, line 119: We are duplicating part of the check in `AuthorizeIssuanceImpl` here, but I guess given we can detect the issue at this point, we can report it earlier as well.

src/main/kotlin/eu/europa/ec/eudi/openid4vci/internal/AuthorizeIssuanceImpl.kt
```diff
@@ -84,8 +88,40 @@ internal class AuthorizeIssuanceImpl(
...
+            AuthorizationResponseIssChecking.Never -> Unit
+
+            AuthorizationResponseIssChecking.IfSupported ->
+                if (metadata.supportsAuthorizationResponseIssuerParam()) {
+                    ensure(!iss.isNullOrEmpty()) { MissingAuthorizationResponseIssuer() }
+                    ensure(issuerMatches(iss, metadata.issuer)) { InvalidAuthorizationIssuer() }
+                }
+
+            AuthorizationResponseIssChecking.Required -> {
+                ensure(metadata.supportsAuthorizationResponseIssuerParam()) {
```
dzarras, line 110: On seconds thoughts, this `ensure` maybe should be replaced by a `check` instead. We are checking for an illegal state here. In `CredentialOfferRequestResolver` we have already verified that the Authorization Server supports RFC9207, and now we expect to receive and validate the `iss` value.

dzarras (changes requested):

src/main/kotlin/eu/europa/ec/eudi/openid4vci/internal/CredentialOfferRequestResolver.kt
babisRoutis: I don't consider this a duplication. It's just a different execution path.

On the one hand, `CredentialOfferResolver` makes sure that if the caller processes a credential offer, we have a fail-fast check.

On the other hand, the similar check on the `AuthorizeIssuance`  interfaces makes sure that no matter the execution path, the requirement to check `iss` is enforced (provided that caller opts for it).

src/main/kotlin/eu/europa/ec/eudi/openid4vci/internal/AuthorizeIssuanceImpl.kt
babisRoutis: I don't agree.

The argument assumes that always we are processing a credential offer. Yet there could be execution paths (or uses of the library if you will) that by pass this step.

src/main/kotlin/eu/europa/ec/eudi/openid4vci/internal/AuthorizeIssuanceImpl.kt
```diff
@@ -84,8 +88,40 @@ internal class AuthorizeIssuanceImpl(
...
         authDetailsOption: AccessTokenOption,
+        issuer: String?,
     ): Result<AuthorizedRequest> = runCatchingCancellable {
         ensure(serverState == state) { InvalidAuthorizationState() }
+        validateAuthorizationResponseIssuer(issuer)
+        requestAccessToken(authorizationCode, authDetailsOption)
+    }
+
+    private fun validateAuthorizationResponseIssuer(iss: String?) {
+        val metadata = credentialOffer.authorizationServerMetadata
```
vafeini, line 99: What if we stored the `iss` value of the AS when preparing the `AuthorizationRequestPrepared`? To me makes more sence to have it there as a state. The `AuthorizationRequestPrepared.authorizeWithAuthorizationCode` will be executed on a prepared state and the passed `issuer` will be compared to that state.

vafeini (approved):

dzarras (approved):

